PR #45, opened 2015-11-15
Changing stpm to grad, since axhm covers stpm.

maxhutch:


maxhutch:
@hfp I see the interfaces are changing, so let's put this on hold until they are stable.  Let me know when they are, and I can merge.

hfp:
I think the interface is rather stable now for Fortran and C++. So, please go ahead -- and sorry for any merge-conflicts you may need to resolve!

However when we change the backend function signatures, things will again change for C code (C is now the weakest language in the chain in terms of supporting certain abstractions [mainly overloading functions]).

About the interface changes: time-wise they were mainly to serve integration into CP2K/trunk, but we always had the feature gap between F and C/C++. The current F interface now validates against F2K3, we got rid of the the reshapes when using the C_LOC-based libxsmm_call signatures, and we now have an equal feature-level with C/C++ (supporting prefetches).

Regarding changing the upcoming backend function signatures (no schedule yet): we will move to (a,b,c) and (a,b,c,x) which is still about having two different signatures (same as now), however our current prefetch signature (a,b,c,pa,pb,pc) cannot be easily supported on Windows. When changing to (a,b,c) + (a,b,c,x), we will only rely on common calling convention between Linux and Windows, and we will also be able passing arbitrary alpha/beta (via x).

maxhutch:
@hfp Updated, merge-ready.

CC: @alheinecke

PR #93, opened 2016-08-11
adds specfem sample

danielpeter:
this PR adds a new sample example "samples/specfem". it mimicks the SPECFEM3D_GLOBE stiffness computation for elastic domains (most heavy routine in the code). 

the example implements 4th-order spectral-element computations (used by default). the various flavors (Deville, unrolled, dispatched, prefetch, static) are timed and speedup compared to the Deville routine (compute_forces_Dev.F90).

a step-by-step explanation is given in the README.md.

best wishes,
daniel


alheinecke:
thanks! That's great that we have a specfem reproducer!

PR #95, opened 2016-08-25
Perf jit profiling support

mdebski:
This pr provides support for perf profiling of the jit code. Both basic (perf map) and extended support (jitdump) is provided:

* Basic is compiled by default if symbols are available (`make SYM=1`), and if enabled in runtime (`LIBXSMM_VERBOSE=-1`, same as for vtune), generates a map file in /tmp, which perf automatically reads and uses to resolve unknown symbols - jitted kernels.
* jitdump can be enabled by using `make SYM=1 PERF_JITDUMP=1`. It generates a "jit-<pid>.dump" file, which includes full information on jitted kernels - addresses, names, sizes, full code dump. This file can be then injected in perf.data (`perf inject -j`) and enables annotated assembly view in perf report in reasonably new perf.

_"jitdump" does not compile currently, as requires `jitdump.h` file from kernel sources (tools/perf/util), which is GPLv2! It works fine after providing this file to src/. I suggest we leave it as-is for now, and try to solve licensing issues independently._


hfp:
Thank you for contributing this facility!
---
I did not know that perf actually gained some support for JIT code. Also, your proposed behavior for "jitdump" looks good to me.

One minor thing (not necessarily an action item) is to implement `libxsmm_perf_write_code` inside of libxsmm_malloc.c (where the VTune support is incorporated). The latter may allow you to get rid of allocating additional data records (as well as calling mmap) within libxsmm_perf.c. Our memory allocation facility (libxsmm_malloc.h and libxsmm_main.h) already allows to attach custom data (which we do in case of executable buffers and VTune support enabled).

PR #98, opened 2016-09-14
Fix dependencies between headers.

mdebski:
Our compilation workflow requires each header to be self-contained - import everything it uses. This achieves it on master by modifying some includes and moving some macros to libxsmm_macros.h and some typedefs to libxsmm_typedefs.h


alheinecke:
Merged fix for includes. This already starts some work we need to do anyway soon.

PR #124, opened 2016-12-16
Bug fix for N that is not a multiple of 16

nrsatish:


jewillco:
This looks to be missing the definition of at least _MM_STOREU_FP32.

PR #135, opened 2017-01-19
Please check line 562 in libxsmm_spmdm.c

nrsatish:


hfp:
I will merge the PR, and then revert the "Potential Bug" as it is correct. I will then continue with 1.6.6 work, so this comes right in time. Anything else let me know by private e-mail.

PR #137, opened 2017-02-10
Weight Update with Padding for NHWC_Custom and NHWC_RSCK

ankushmandal:
I have added support in Weight Update kernel for NHWC_custom and NHWC_RSCK format. Currently it is giving correct result for convolutions with padding and without padding.


alheinecke:
Can you please fix the CI issues, then we can merge

hfp:
This is about C89 conformance or mixed declaration and code: https://travis-ci.org/hfp/libxsmm/jobs/200193853#L3393 (and following lines).

Moving variable declarations into the leading block is the typical solution, but if it's destroying locality of the code one can also introduce an additional scope {} (curly braces), and start (additional) variable declarations inside).

alheinecke:
can you please remove src/generator_convolution_weight_update_avx512.c.backup  from the changeset?

PR #138, opened 2017-02-10
Fixed run_deepbench.sh to run only libxsmm format

dmudiger:


alheinecke:
Can we add a new script which only runs "A L", right now the script doesn't support the other formats anymore, but within a couple of we will have a performance fix (c.f. #137)

dmudiger:
For DeepBench we would want only the best performing variant. I can modify the script to take the format as an argument.

alheinecke:
That would be great, perhaps we should also add the last parameter to toggle logical vs, physical padding?!

alheinecke:
thanks and merged.

dmudiger:
done, updated run script with additional arguments.

PR #139, opened 2017-02-13
Added support for KNM for Convolution using Winograd

kunalbanerjee:
Modified convolution using winograd to support KNM architecture


alheinecke:
Thanks! Merged!

PR #147, opened 2017-03-01
Winograd knm

kunalbanerjee:
Added non-NULL checks for scratch memory, checks for divisibility by 4 for qfma and fixed some code to do away with FDVLEN and VRATIO


alheinecke:
Can you please resolve the conflicts with the master and resubmit?

PR #154, opened 2017-03-22
Backward convolution code performance improvement

ankushmandal:
The new backward code structure is similar to forward code. Furthermore, I have included possible optimizations specific to backward code.


hfp:
Good work!  I've taken all changes, but not the "old" files. We can keep some amount of dead code, but not at file level. Of course, one can do this kind "versioning" (or "todolist") locally (outside of the repository).

PR #165, opened 2017-05-04
Fixed casting and mask move instruction

egeor:


hfp:
I revised the code further in 29445bd just to make it clear that the mask is taken as an integer.

PR #176, opened 2017-05-31
Winograd knm

kunalbanerjee:
Unrolled along qfma for convolutions using Winograd


hfp:
It fails right now, but I will take it locally, fix, and push it -- don't worry!

PR #178, opened 2017-06-19
modified check for pad  applicable to googlenet v3

taknevski:


alheinecke:
Can you please also submit it a script which tests Googlenetv3 convolutions in samples/dnn

taknevski:
Hi Alex,

I can do that, but the reproduced needs an additional field for pad_w
because pad_w may not equal pad_h in googlenet v3

Regards,
Anand

hfp:
Hello Anand,

perhaps  e5bac4b will help to get a distinct `pad_w`; feel free to adjust until it suits the requirement.

Hans

alheinecke:
I think the problem is that we need pad cmd parameter in H and in W direction. So we need to change all the scripts etc. However, this is what needs to be done. 
Today we have:
./layer_example_f32 iters inpWidth inpHeight nImg nIfm nOfm kw kh pad stride type format padding_mode
and we need:
./layer_example_f32 iters inpWidth inpHeight nImg nIfm nOfm kw kh padw padh stride type format padding_mode
For all our current scripts: padw=padh

taknevski:
yes you need an addiitonal parameter for pad_w..

alheinecke:
Can you please include the driver changes for fp32, int16, int8 and the modification of all other scripts into this PR?

Thanks!

hfp:
How to proceed with this PR, is there some more work pending?

PR #179, opened 2017-06-20
Added optimized BGEMM routine

dmudiger:
Added optimized BGEMM routine, with M,N,K-way parallelism (replaced samples/blkgemm)
Contributors: @kunalbanerjee


alheinecke:
great! Thanks!

Can you please apply to following changes for a better integation:
- split it into src/libxsmm_blkgemm.c and samples/blkgemm (driver and implementation is src)
- rename all functions to libxsmm_* -> this is our C-namespace
- providing a run script for DeepBench

alheinecke:
thanks!

Unit tests fail with:
/home/travis/build/hfp/libxsmm/samples/blkgemm/block_gemm.c:42:10: fatal error: 'libxsmm_blkgemm.h' file not found

#include "libxsmm_blkgemm.h"
Can you please have look?

hfp:
As soon as compilation passes, I can take over, merge locally, brush it up further, and merge publicly - no need for you guys to match all the pedantic requirements. I will also look into integration with the interface used for the current tile-GEMM implementation.

alheinecke:
Great! Thanks!

alheinecke:
still seems to fail:

make: *** No rule to make target `/home/travis/build/hfp/libxsmm/include/libxsmm_blkgemm.h', needed by `include/libxsmm.h'.  Stop.

dmudiger:
Don't see that error with icc ?

hfp:
Minor comment: the way the header is included is kind of wrong or lazy: header files from a search path (-I) should be included using angle-brackets, whereas double-quotes are for relative paths.

dmudiger:
Yes, I spotted that and it also seems like there are few more remnants of the cean usage.
@kunalbanerjee - can you please look into this, make sure they are changed to the VLA semantics and push it back in. thanks.

hfp:
I briefly looked over the code and have a number of comments:
* The code/header uses leading underscores for macros (or functions), which is non-conforming for user-level functions since leading underscores are reserved for compiler builtins.
* The code/header uses unprefixed/common symbol names (macros, functions, etc.) in the global namespace, which not only pollutes the global namespace but can cause problems with user code (name clash). For instance, using "real" as a symbols name is an awful idea.
* The code duplicates functionality, which is wrapped up already: synchronization primitives (see libxsmm_sync.h), macro injection aka _Pragma (see libxsmm_macros.h). There is a good reason to rely on existing code: portability. It not sufficient if this code passes on a Linux box. We support OSX (implies BSD somewhat), all sorts of Linux, and the code must compile and work on native Microsoft Windows.
* If something is missing on our abstraction layer (sync, or whatever) - please feel free to introduce it carefully in one of these places.
* Please create header, which only introduces *our* interface (without including tons of system headers, or even missing to include). For example, having mmap in our public interface is not a good idea. Please do not use inline functions at the interface level (public header) unless really needed.
* Please move implementation helper macros and such into the translation unit, or introduce a private header file (src folder) if you need to reuse such helpers across multiple translation units.
* Since the code uses OpenMP, it should not go into main LIBXSMM, but rather libxsmmext. Our main library is agnostic wrt a threading runtime.
* Using `extern "C"` in the header is not proper; this is already managed by decorating functions with LIBXSMM_API, etc.
* There are C++-style comments (//), which will not pass with C89.
I believe the code is relatively far from passing with our Travis tests. I can take the code later today. If you want to continue, please send me a PM to stop me from duplicating work.

alheinecke:
I think we should merge into the main part of the library by making the interface identical to convolutions and spdm (with appended _st). in extlibxsmm we can provide an OpenMP wrapped version.

hfp:
I may be able to have it integrated by Friday EOD (my time). However, completion may slip into next week.

alheinecke:
Great! Thanks!

PR #189, opened 2017-08-08
Winograd

kunalbanerjee:
Removed redundant code for Winograd; removed "avx512" from the file names; added non-overwrite cases for Jongsoo's code.


hfp:
Right now, all these internal functions (e.g., internal_bwd_input_transform_custom_custom) are not called anymore (dead code). Previously these functions implemented the CPUID-dispatch, and we had always two code paths: (1) intrinsic based AVX-512F, and (2) non-intrinsic code path (which was compiled according to the target flags given on the compiler's command line). We compiled functions belonging to path (1) using LIBXSMM_INTRINSICS(LIBXSMM_X86_AVX512). This decorated the function such that the compiler was able to generate AVX-512 code. Unlike Intel Compiler, the GCC and Clang compilers *need* target flags on the command line ("static code path"), or as a target-attribute decorating a specific function.

I do not understand how it is supposed to work now. Btw, you *cannot* use for instance LIBXSMM_DNN_CONVOLUTION_WINOGRAD_BACKWARD_AVX512 to decide if something will be compiled for AVX-512 ("static code path"). The meaning of this definition is the following: "dear compiler, are you able to generate this kind of code because you understand the target-attribute.". Also, using the previous CPUID-dispatch mechanism we were able to run AVX-512 code even when the whole library was compiled with something lower than AVX-512. The whole mechanism to check whether the compiler is able to do such "magic" is abstracted in libxsmm_intrinsics.h. Sorry, kind of complicated ;-)

hfp:
I guess this PR can be closed?

kunalbanerjee:
Yes

Regards,
Kunal Banerjee
Research Scientist
Intel Parallel Computing Lab
Bangalore, India
Mob: +91 9432255217

PR #192, opened 2017-08-10
Winograd with default

kunalbanerjee:
Fixed jongsoo's code to support overwrite and non-overwrite cases; also got rid of reuseInput option because fwd/bwd and upd have different layouts


hfp:
>>> got rid of reuseInput option

That's really great! This piece was an awful hack anyways with no proper interface (it toggled something inside of the library; pretty bad).

hfp:
Superceeded by another PR.

PR #224, opened 2018-05-30
Updated layout code for rnn and lstm

kunalbanerjee:


hfp:
I believe, the helper-header should be either private (i.e., only in the `src/` directory) or it may be something limited to the RNN/LSTM sample code. I will leave the review up to Alex since I am not aware of all intents. If the helper-header is indeed meant for `include`, a more conforming filename would be appreciated e.g., `libxsmm_dnn_recurrent.h` - but as I said, I have not looked at the details.

alheinecke:
I agree, we need to move that into the internal part. I guess (files are missing) that these are the element-wise routines? I suggest for now having them in "libxsmm_dnn_elementwise.c/h", both located in src/.

hfp:
Please merge hfp/master as soon as possible (to avoid potential conflicts) due to some changes on top of this work.

PR #230, opened 2018-07-09
Added bwd/upd passes for RNN for both correctness and performance

kunalbanerjee:


hfp:
At least template/libxsmm_bgemm_convert_b_to_a.tpl.c is missing (git add).

Issue #276, opened 2018-10-30
[pkg-config] Add required libs/linker-flags to Libs for static-only build


dev-zero:
If only the static library is built, then should the linker flags and required libraries be added to the main `Libs` entry instead of the `Libs.private`, see also https://github.com/mesonbuild/meson/pull/3093
The problem is that it is otherwise a guess-work for the consumer whether or not he has to use the static flags.

PR #278, opened 2018-10-30
pkg-config: add private libs to main libs for static-only build

dev-zero:
fixes: #276

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1714,6 +1714,20 @@ ifneq ($(abspath $(INSTALL_ROOT)),$(abspath .))
 	@$(CP) -v .state $(INSTALL_ROOT)/$(PDOCDIR)/artifacts/make.txt
 endif
 
+ifeq (Windows_NT,$(UNAME))
+  XSMM_PKG_CONFIG_LIBS_PRIVATE = "-ldbghelp"
+else ifneq (Darwin,$(UNAME))
+  ifneq (FreeBSD,$(UNAME))
+    XSMM_PKG_CONFIG_LIBS_PRIVATE = "-lpthread -lrt -ldl -lm -lc"
+  else
+    XSMM_PKG_CONFIG_LIBS_PRIVATE = "-ldl -lm -lc"
+  endif
+endif
+
+ifneq (Darwin,$(UNAME))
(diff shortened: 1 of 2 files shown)


hfp:
Hmm, I wonder if `pkg-config --static` was exactly meant to spit out the link-line for the static case. With this patch, the generated pc-files become specific/dependent on STATIC=0|1.

hfp:
For me (without the patch) the pkg-config stuff looks like:

```bash
PKG_CONFIG_PATH=/path/to/libxsmm/lib pkg-config libxsmm --libs
-L/path/to/libxsmm/lib -lxsmm
```
and for the static link-line:

```bash
PKG_CONFIG_PATH=/path/to/libxsmm/lib pkg-config libxsmm --libs --static
-L/path/to/libxsmm/lib -lxsmm -lpthread -lrt -ldl -lm -lc
```

Is this not what you wanted?

hfp:
I understand what you would like to achieve:

* Pickup static library (if it is present); link-line incluses "private libraries".
* Pickup shared library if no static library is available.

However, for instance `SHARED` is not what is seems to be (though STATIC=0 is already "shared"). In fact, SHARED is only a handy key to trigger building static as well as shared libraries (was/is needed for the Debian package deployment). Anyhow, what people eventually do is building LIBXSMM twice (`make` and `make STATIC=0`; and the order is people's choice). The logic proposed by the patch cannot infer what's needed (check existence of static library) because people can first execute either `make` or `make STATIC=0`; it makes the default link-line suggested by pkg-config unreliable.

I will refine the patch and try to achieve what you propose.

dev-zero:
no, the order is quiet the other way round (matching the order in which compilers are resolving the `-lxsmm` in the absence of `-static`):

* Pickup shared library if available (`pkg-config --libs libxsmm` should give just `-L... -lxsmm`, `pkg-config --libs --static libxsmm` should give `-L... -lxsmm -ldl -lrt ...` if a static library is also available, otherwise the same as without `--static`)
* Pickup static library if no shared library available (`pkg-config --libs libxsmm` should give `-L... -lxsmm -ldl -lrt`, `pkg-config --libs --static libxsmm` should give the same thing)

PR #293, opened 2018-12-10
Optimize rnn alternative batchreduce gemm

egeor:


hfp:
I have merged this PR into hfp/develop with some fixups on top. However, the code uses Variable Length Arrays (VLAs) -- this must be reworked to use dynamically allocated memory from somwehere (handle). To treat such buffers as multi-dimensional array, one can rely on our VLA-macros (`LIBXSMM_VLA_DECL` and `LIBXSMM_VLA_ACCESS`).

PR #304, opened 2019-01-14
Gxmupd2

ska278:


alheinecke:
Can you please double checkout the changes in LIBXSMM and if the code still converges when reverting them:
a) eps back to 1e-7
b) double precision should be replaced by single precision again

ska278:
Convergence confirmed with eps=1e-7 and single-precision mean/variance calculation

PR #307, opened 2019-01-16
lstmdriver: Introduced a parallel for loop to compute delbias in refe…

kunalbanerjee:
…rence code


hfp:
The `for (int p` is C99 and needs to be changed such that `int p` is declared upfront (OMP-private needs to be updated then as well).

hfp:
Btw, not having C99 or only C90 aka C89 is not bug.

PR #320, opened 2019-02-15
Initial changes for pulling grucell into rnncell

kunalbanerjee:


kunalbanerjee:
Please note that grudriver is non-functional at the moment; I shall fix it soon (this is only the initial changes that are required).

PR #324, opened 2019-03-21
Fixed gru bwd/upd for cases C==K and K%2048==0

kunalbanerjee:


hfp:
I will take the PR, but can you please pull and double-check `rnndriver_nc_ck.c:324`?

In your original code `j` was not initialized, but even if it is initialized (as I did) it looks somewhat cumbersome i.e., the purpose of `j` is not fully clear to me.

PR #362, opened 2019-12-14
Newfullyconnected

kunalbanerjee:
Current code compiles and runs correctly.
However, in src/libxsmm_dnn_fullyconnected_backward.c if I comment line 110 and uncomment 109 - it doesn't compile. Need help to figure out why.


hfp:
In `libxsmm_dnn_fullyconnected_st_bwd_ncnc_kcck_generic.tpl.c:81` you attempt to call your kernel as defined in line 109. The call signature does not match: `A_array` and `B_array` consist of `unsigned long long` however, `libxsmm_typedefs.h` defines the type `libxsmm_bsmmfunction_reducebatch_addr` to take arrays of pointers to `libxsmm_bfloat16`.

Make sure you compile also with `PEDANTIC=2` to get no surprise/complaint from Travis CI.

alheinecke:
What is the performance benefit? Is it now the same as forward?

kunalbanerjee:
> What is the performance benefit? Is it now the same as forward?

fullyconnected fwd performed best with stride_brgemm; currently only address_brgemm is implemented for bwd and upd -- so, there's scope of some improvement -- however, have to get the correctness fixed first

hfp:
If I look at Travis' output, it appears that unresolved merge conflicts ended up somehow in this PR.

Issue #378, opened 2020-04-19
A Sparse CSR Fallback Kernel


mahudu97:
When using `libxsmm_dfsspmdm_create` to JIT a kernel for sparse A, if the number of unique constants does not fit in a register kernel, the fallback is to use a dense routine.

Is there a reason as to why `libxsmm_generator_spgemm_csr_asparse` is not considered?

A GitHub search shows that `libxsmm_generator_spgemm_csr_asparse`  \ `libxsmm_generator_spgemm_csr_kernel` is unused throughout the library.

alheinecke:
We offer "soa" and "packed" kernels. The simple CSR kernel where never needed in recent past and the intrinisic/static kernel have been enough. 

There are plans to pack more values into the kernel, but the benefits so far have been minor. What is you concrete application?

mahudu97:
The use case is the sparse matrix multiplication that occurs in `PyFR`.

I have a method that packs more values into the simple CSR kernel, in my fork. It's part of my thesis (Imperial College, Supervisor Prof. Paul Kelly). Link: https://github.com/mahudu97/libxsmm/commit/815d80bab476a5e884190c9f5be74011b5ba0b7d
The intention is to make a pull request after evaluating within my thesis. In the PR I'd of course provide a detailed explanation of the method / changes.
In summary, it packs up to `120` unique elements in an AVX-512 register file (both DP and SP), and broadcasts with one instruction (3 cycle latency) regardless of the lane the element is stored in.

I had a look at the "soa" kernels. I assume "soa" is Struct of Arrays? (Sorry not too familiar with some of the matrix terminology, if its something else).
I noticed it uses `VBROADCASTSX` as its move `A` instruction. Does this kernel load the values of `A` from memory? (as opposed to broadcasting values already in registers, as I believe only lane zero can be broadcast using this instruction)

Finally, I'm not too sure what the `packed` family of GEMM generators are used for, could you please give me an indication of their use? Thanks.

alheinecke:
Ah, great, yes I talked with Freddie and Paul ~8months ago and that we should write such a kernel:-) We are happy to take the changes. When Freddie and Paul and I talked we had 3 kernel variants in discussions:
o	RegA -> with permutes
o	RegA -> with Shuffles
o	RegA -> with compact A storage and loading form L1$
I think your code is the second case, correct?
[...]

PR #381, opened 2020-05-12
Issue #378: Add run-time broadcasting to csr_asparse_reg using permutes

mahudu97:
Hybrid Approach to register based routine.

If over `31` unique non-zeros, pack values then broadcast with permute.
DP supports up to 176 and SP supports up to 224.

Issue #378 details high-level changes in more detail.


alheinecke:
Many thanks!

hfp:
Thank you for your contribution!

@alheinecke Let´s add some prefix (`LIBXSMM_`) to

```c
#define BCAST_REG 30
#define ACC_REG 31
#define MAX_UNIQUE_DP 176
```

... and similar defines. ( Undefining at the bottom of the translation unit is probably a pattern reserved for our "templates". )

Issue #393, opened 2020-07-24
xsmm should migrate to using Eigen::bfloat16


tjleeyon:
bfloat16 is now part of Eigen; xsmm should migrate to using Eigen::bfloat16 instead of tensorflow::bfloat16.

hfp:
Task: attempt to rely on zero-config (TF Bazel/build file for LIBXSMM).

hfp:
I knew our integration with TF is essentially EOL and outdated. In fact, TF/master does not build with XSMM-knobs (since we revised our Deep Learning API since then). I may fix this aka "my last effort" (probably not). Potentially PlaidML's (plaidml-v1 branch) can revive TF+LIBXSMM even in a more thorough fashion (due to a TF-bridge being part of PlaidML).

@tjleeyon Are you interested in integrating Eigen and LIBXSMM? Or let me ask: what is your motivation for this PR and how can we potentially help?

alheinecke:
Can we close this issue?

hfp:
This issue is not current or not applicable anymore.

PR #394, opened 2020-07-24
Switch to using Eigen::bfloat16 instead of tensorflow::bfloat16.

tjleeyon:
This is to address issue #393 since tensorflow will move to using Eigen::bfloat16 directly instead of defining its own struct, which will break xsmm when building together with tensorflow code without this change.


hfp:
Thank you for your contribution!

PR #414, opened 2020-09-21
[WIP] Add sparse training

ypsoh:
- [x] Add readme.md
    - [x] Data preparation steps
- [x] Add training scripts
    - [x] ResNet
    - [x] Transformers
    - [x] DLRM


hfp:
Would it be possible to cleanup the new content such that our [archive size](https://github.com/hfp/libxsmm/archive/master.zip) drops into previous region?

It almost tripled due to this series additions, and I noticed, e.g., `tmp1` directory (archives, prebuilt Python wheels), `bak` files, and documentation (incl. large animated `gif` file). For the documentation, it can be an option to refer to the original project (only include prebuilt documentation if this is original to your contribution).

ypsoh:
Hi, Thanks. I'll try to take the size down a bit. Is there a size I should be targeting? I'm not quite sure what you mean by `drops into previous region`.

hfp:
Let´s not overdo this, I really do not want to command your time. It is just to get your help rather than me flushing out the wrong bits. To answer your question: for instance, a [revision](https://github.com/hfp/libxsmm/archive/61528b38a411763b69325e6837f44505afdae730.zip) prior to your work (but after 1.16.1) is ~3.3 MB in size (btw, v1.16.1 was about 3.4 MB). Current size of the zip-archive is 8.6 MB. So, I wonder if some of the content mentioned in my earlier comment can be just mentioned as URL (documentation and wheels), and if some files are actually temporary data. If nothing applies we will be up to 9 MB; fine too.

PR #419, opened 2020-10-21
Remove files that take up too much space in sparse_training sample folder

ypsoh:
This commit reduces the size of the `samples/sparse_training` folder and the overall size of the repository


ypsoh:
In response to: https://github.com/hfp/libxsmm/pull/414#issuecomment-704218338 @hfp

hfp:
Thank you very much!

PR #429, opened 2020-11-24
1D dilated convolutional layer. 

narendrachaudhary51:
This folder contains a PyTorch convolutional layer that can be used to replace the existing PyTorch nn.Conv1d layer. We wrote the optimized convolutional layer code using libxsmm library's brgemm functions. This optimized layer has both float32 and bfloat16 implementations.


alheinecke (changes requested):
This file has no lic headers, where is the code from?

Why are the LIBXSMM intrinsics and/or TPPs for FP32<->BF16 not used. I suggest using the LIBXSMM TPPs (mateltwise) for this

samples/deeplearning/conv1dopti_layer/Conv1dOpti-extension/Bfloat16.h
```diff
@@ -0,0 +1,42 @@
+/* create a type called bfloat16 – need to figure out a common header file */
```
alheinecke, line 1: This file has no lic headers, where is the code from?

Why are the LIBXSMM intrinsics and/or TPPs for FP32<->BF16 not used. I suggest using the LIBXSMM TPPs (mateltwise) for this

samples/deeplearning/conv1dopti_layer/Conv1dOpti-extension/Conv1dOpti.cpp
```diff
@@ -0,0 +1,745 @@
+#include<immintrin.h>
```
alheinecke, line 1: Please add lic header

samples/deeplearning/conv1dopti_layer/Conv1dOpti-extension/Conv1dOpti.cpp
```diff
@@ -0,0 +1,745 @@
...
+// #include<x86intrin.h>
+#include<iostream>
+#include<stdio.h>
+// #include<stdlib.h>
+#include<torch/extension.h>
+#include<tuple>
+#include<omp.h>
+#include<libxsmm.h>
+#include<libxsmm_intrinsics_x86.h>
```
alheinecke, line 10: see comment on BF16 intrinsic above

samples/deeplearning/conv1dopti_layer/Conv1dOpti-extension/Conv1dOpti.cpp
```diff
@@ -0,0 +1,745 @@
...
+#define XS_TILE_DBACKWARD 64
+#define XS_TILE_WBACKWARD 256
+
+
+
+LIBXSMM_API_INLINE LIBXSMM_INTRINSICS(LIBXSMM_X86_AVX512_CORE)
+void bf16_vnni_reformat(libxsmm_bfloat16 *_in, libxsmm_bfloat16 *_out, int M, int N, int ld_in, int ld_out) {
+    /* Function to do VNNI transform */
+
+#if defined(LIBXSMM_INTRINSICS_AVX512_CORE)
```
alheinecke, line 28: here we have the mateltwise transform TPP for VNNI formating

samples/deeplearning/conv1dopti_layer/Conv1dOpti-extension/Conv1dOpti.cpp
```diff
@@ -0,0 +1,745 @@
...
+    libxsmm_xtransfunction tr_kernel;
+    libxsmm_descriptor_blob blob;
+    libxsmm_trans_descriptor* tr_desc;
+    tr_desc = libxsmm_trans_descriptor_init(&blob, typesize, M, N, LDO);
+    tr_kernel = libxsmm_dispatch_trans(tr_desc);
+    PCL_ASSERT(tr_kernel, "Null Transpose kernel");
+    return tr_kernel;
+}
+
+void convert_f32_bf16(float* in, bfloat16* out, int len)
```
alheinecke, line 88: here we also have a mateltwise TPP

samples/deeplearning/conv1dopti_layer/Conv1dOpti-extension/Conv1dOpti.cpp
```diff
@@ -0,0 +1,745 @@
...
+
+    /* Dispatch brGEMM kernels for the normal case and the edge case*/
+    libxsmm_bmmfunction_reducebatch_strd bmmshortkernel = libxsmm_bmmdispatch_reducebatch_strd(XS_TILE_FORWARD, F_t, C_t, dial*2*sizeof(libxsmm_bfloat16), F_t*C_t*sizeof(libxsmm_bfloat16), &short_width, &lda, &ldc, NULL, NULL, NULL, NULL);
+    libxsmm_bmmfunction_reducebatch_strd bmmedgekernel2 = libxsmm_bmmdispatch_reducebatch_strd(W_t - tile_multiple, F_t, C_t, dial*2*sizeof(libxsmm_bfloat16), F_t*C_t*sizeof(libxsmm_bfloat16), &edge_width, &lda, &ldc, NULL, NULL, NULL, NULL);
+
+    #pragma omp parallel for
+    for(int n = 0; n < N_t; n++) {                               // Loop for batches
+        int last_block = 0;
+        for(int wb = 0; wb < W_t - XS_TILE_FORWARD + 1; wb += XS_TILE_FORWARD) {    // width blocking loop (Normal case)
+            for(int filter = 0; filter < F_t; filter++){                            /* Loop for initilization of output array */
```
alheinecke, line 169: we have a matetlwise kernel for setting 2d arrays to 0

samples/deeplearning/conv1dopti_layer/Conv1dOpti-extension/Conv1dOpti.cpp
```diff
@@ -0,0 +1,745 @@
...
+    auto Y = input.new_empty({N_t,F_t,W_t});        // New tensor for output
+
+    libxsmm_bfloat16* input_a = (libxsmm_bfloat16*) input.data_ptr<at::BFloat16>();       // Get BFloat16 data pointers for accessing tensors
+    libxsmm_bfloat16* weight_a = (libxsmm_bfloat16*) weight.data_ptr<at::BFloat16>();
+    libxsmm_bfloat16* Y_a = (libxsmm_bfloat16*) Y.data_ptr<at::BFloat16>();
+
+    auto flip_weight = weight.new_empty({WW_t,F_t,C_t});                                // Weight tensor with permuted dimension (width, filters, channels)
+    libxsmm_bfloat16* flip_weight_a = (libxsmm_bfloat16*) flip_weight.data_ptr<at::BFloat16>();    // Get BFloat16 data pointers for accessing the tensor
+
+    for(int kw = 0; kw < WW_t; kw++){                               // Loop to permute weight tensor dimensions
```
alheinecke, line 135: Could a transpose TPP used here, is this a common kernel?

samples/deeplearning/conv1dopti_layer/Conv1dOpti-extension/Conv1dOpti.cpp
```diff
@@ -0,0 +1,745 @@
...
+
+    int pad_tile_multiple = 2 * (((WW_t - 1)*dial)/XS_TILE_DBACKWARD + 1) * XS_TILE_DBACKWARD;       // Padded block/tile (896)
+    int ldb_shortpad = 2*pad_tile_multiple;       // grad padded short buffer (1792)
+
+    auto grad_shortpad_tensor = grad.new_empty({N_t,F_t,2*pad_tile_multiple});
+    libxsmm_bfloat16* grad_a_shortpad = (libxsmm_bfloat16*) grad_shortpad_tensor.data_ptr<at::BFloat16>();   // short buffer for padded gradiant
+
+    #pragma omp parallel for
+    for(int n = 0; n < N_t; n++){                   // loop to store the edges for gradiant array into grad_a_shortpad buffer
+        for(int filter=0; filter < F_t; filter++){
```
alheinecke, line 253: could we use a mateltwise kernel here? Do we need to add one?

samples/deeplearning/conv1dopti_layer/Conv1dOpti-extension/Conv1dOpti.cpp
```diff
@@ -0,0 +1,745 @@
...
+    /* Dispatch brGEMM kernels for the normal case and the edge case*/
+    libxsmm_bmmfunction_reducebatch_strd bmmshortkernel = libxsmm_bmmdispatch_reducebatch_strd(XS_TILE_DBACKWARD, C_t, F_t, 2*dial*sizeof(libxsmm_bfloat16), C_t*F_t*sizeof(libxsmm_bfloat16), &short_width, &lda, &ldc, NULL, NULL, NULL, NULL);
+    libxsmm_bmmfunction_reducebatch_strd bmmshortkernel2 = libxsmm_bmmdispatch_reducebatch_strd(Win_t - tile_multiple, C_t, F_t, 2*dial*sizeof(libxsmm_bfloat16), C_t*F_t*sizeof(libxsmm_bfloat16), &short_width, &lda, &ldc, NULL, NULL, NULL, NULL);
+
+    #pragma omp parallel for
+    for(int n = 0; n < N_t; n++) {
+        int last_block=0;
+
+        for(int wb = 0; wb < Win_t - XS_TILE_DBACKWARD + 1; wb += XS_TILE_DBACKWARD) {
+            for(int channel = 0; channel < C_t; channel++){                 /* Loop for initilization of grad array */
```
alheinecke, line 289: we have a mateltwise kernel for setting a matrix to 0

samples/deeplearning/conv1dopti_layer/Conv1dOpti-extension/Conv1dOpti.cpp
```diff
@@ -0,0 +1,745 @@
...
+    int short_W_t = XS_TILE_WBACKWARD;
+    int edge_W_t = W_t - tile_multiple;
+
+    auto grad_shortvnni_tensor2 = grad.new_empty({N_t,F_t,short_W_t});                            // Short buffer for storing VNNI transform
+    libxsmm_bfloat16* grad_shortvnni = (libxsmm_bfloat16*) grad_shortvnni_tensor2.data_ptr<at::BFloat16>();
+
+    auto grad_edgevnni_tensor2 = grad.new_empty({N_t,F_t,edge_W_t});                              // Short buffer for storing VNNI transform in edge case
+    libxsmm_bfloat16* grad_edgevnni = (libxsmm_bfloat16*) grad_edgevnni_tensor2.data_ptr<at::BFloat16>();
+
+    libxsmm_xtransfunction trans_shortkernel_grad = get_tr_kernel(sizeof(float), short_W_t/2, F_t, F_t);      // Dispatch transpose kernel
```
alheinecke, line 363: it's better to use the mateltwise transform kernel

samples/deeplearning/conv1dopti_layer/Conv1dOpti-extension/Conv1dOpti.cpp
```diff
@@ -0,0 +1,745 @@
...
+    float* d_input_a = d_input.data_ptr<float>();
+    float* d_weight_a = d_weight.data_ptr<float>();
+
+    /*  Backward data part of the code */
+
+    auto flip_weight = weight.new_empty({WW_t,C_t,F_t});                  // Tensor for permuted weights (width, channels, filters)
+    float* flip_weight_a = flip_weight.data_ptr<float>();
+
+    for(int kw = 0; kw < WW_t; kw++){                   // Loop to permute and flip weights
+        for (int channel=0; channel < C_t; channel++){
```
alheinecke, line 518: can we use a  mateltwise kernel?

samples/deeplearning/conv1dopti_layer/Conv1dOpti-extension/Conv1dOpti.cpp
```diff
@@ -0,0 +1,745 @@
...
+        if(input_a[w] >= 32768)                         // sign bit indicates if value is negative
+            Y_a[w] = 0;
+        else
+            Y_a[w] = input_a[w];
+    }
+
+    return Y;
+}
+
+at::Tensor relu_backward_bf16(at::Tensor& grad, at::Tensor& output){
```
alheinecke, line 709: why not using a framework operation here?

samples/deeplearning/conv1dopti_layer/Conv1dOpti-extension/Conv1dOpti.cpp
```diff
@@ -0,0 +1,745 @@
...
+                d_weight_a[filter*C_t*WW_t + channel*WW_t + kw] = flip_d_weight_a[kw*C_t*F_t + channel*F_t + filter];
+            }
+        }
+    }
+
+    return {d_input, d_weight};         // return data gradiant and weight gradiant
+}
+
+
+at::Tensor relu_forward_bf16(at::Tensor& input){
```
alheinecke, line 686: why not using a framework operation here?

samples/deeplearning/conv1dopti_layer/Conv1dOpti-extension/build/temp.linux-x86_64-3.6/.ninja_log
```diff
@@ -0,0 +1,4 @@
+# ninja log v5
```
alheinecke, line 1: This file should not be part of the PR as it's not source.

samples/deeplearning/conv1dopti_layer/Conv1dOpti-extension/build/temp.linux-x86_64-3.6/build.ninja
```diff
@@ -0,0 +1,20 @@
+ninja_required_version = 1.3
```
alheinecke, line 1: This file should not be part of the PR as it's not source.

samples/deeplearning/conv1dopti_layer/Conv1dOpti-extension/conv1dopti_layer.egg-info/PKG-INFO
```diff
@@ -0,0 +1,10 @@
+Metadata-Version: 1.0
```
alheinecke, line 1: This file should not be part of the PR as it's not source.

alheinecke (changes requested):
There are several issues and also binary and log files are part of the PR which should be removed

samples/deeplearning/conv1dopti_layer/Conv1dOpti-extension/Conv1dOpti.cpp
ddkalamk: Pytorch still doesn't support ReLU for BF16... So we need this until they add it

alheinecke: @egeor I think we can generate a standalone TPP for ReLU in BF16 correct?

alheinecke: https://github.com/hfp/libxsmm/blob/master/include/libxsmm_typedefs.h#L179

egeor: > @egeor I think we can generate a standalone TPP for ReLU in BF16 correct?

Sure. In fact I have already a TPP for bwd relu, however it reads bitmasks. it can be trivially extended to read bf16 tensor.

samples/deeplearning/conv1dopti_layer/Conv1dOpti-extension/Conv1dOpti.cpp
narendrachaudhary51: I modified the existing function (replace line 54 by line 55). I needed to do that to handle VNNI reformat for submatrix. I mentioned this to @egeor. He said that we should not make the change in existing function and suggested me to make the changes locally in my code.

alheinecke: hmm: this code should handle this: https://github.com/hfp/libxsmm/blob/master/samples/eltwise/eltwise_transform.c#L371 or what do you need differently?

narendrachaudhary51: I need to use different values for ld_in and ld_out. I can't use the same value for both. I don't see ld_in and ld_out options on this function.

alheinecke: https://github.com/hfp/libxsmm/blob/master/samples/eltwise/eltwise_transform.c#L424

samples/deeplearning/conv1dopti_layer/Conv1dOpti-extension/conv1dopti_layer.egg-info/PKG-INFO
narendrachaudhary51: Removed this file

samples/deeplearning/conv1dopti_layer/Conv1dOpti-extension/build/temp.linux-x86_64-3.6/build.ninja
narendrachaudhary51: Removed this file.

samples/deeplearning/conv1dopti_layer/Conv1dOpti-extension/build/temp.linux-x86_64-3.6/.ninja_log
narendrachaudhary51: Removed this file.

samples/deeplearning/conv1dopti_layer/Conv1dOpti-extension/Conv1dOpti.cpp
narendrachaudhary51: Added lic header.

samples/deeplearning/conv1dopti_layer/Conv1dOpti-extension/Conv1dOpti.cpp
narendrachaudhary51: In the FP32 version of this code. I replace the existing transform kernel with libxsmm_dispatch_meltw_transform() and I am observing 3-4% slowdown. Is this expected? I am taking a lot of small matrix transforms in loop, instead of one large matrix transform.

samples/deeplearning/conv1dopti_layer/Conv1dOpti-extension/Conv1dOpti.cpp
narendrachaudhary51: Same as the forward pass kernel. The PyTorch framework doesn't support ReLU for BF16.

samples/deeplearning/conv1dopti_layer/Conv1dOpti-extension/Conv1dOpti.cpp
narendrachaudhary51: In this backward pass implementation, padded version of the output gradient tensor is needed to handle some edge cases. We are copying the essential information into a smaller array that stores only edge+padding information of the padded gradient array. This way, we don't need to copy and use the whole array during compute. This reduces the bandwidth requirement and speeds up the code. I don't know how common this operation is to have an TPP kernel of it's own.

samples/deeplearning/conv1dopti_layer/Conv1dOpti-extension/Bfloat16.h
alheinecke: Why are these ops still used and not the intrinsiscs which come with LIBXSMM

samples/deeplearning/conv1dopti_layer/run.sh
```diff
@@ -6,4 +18,4 @@ export KMP_AFFINITY=compact,1,0,granularity=fine              # Set KMP affinity
 export OMP_NUM_THREADS=28                                     # Set number of threads
 export LD_LIBRARY_PATH=/nfs_home/nchaudh1/libxsmm/lib/        # Set LD_LIBRARY_PATH
```
alheinecke, line 19: Please make the path relative

samples/deeplearning/conv1dopti_layer/Conv1dOpti-extension/Conv1dOpti.cpp
alheinecke: But what about the LIBXSMM TPPs?

samples/deeplearning/conv1dopti_layer/Conv1dOpti-extension/Conv1dOpti.cpp
alheinecke: What are the sizes? Are they falling into the optimized path of the transposes?

samples/deeplearning/conv1dopti_layer/Conv1dOpti-extension/Conv1dOpti.cpp
alheinecke: Do you have a more detailed explanation what is needed in an abstract way then we can decide if a TPP is good.

samples/deeplearning/conv1dopti_layer/Conv1dOpti-extension/Bfloat16.h
narendrachaudhary51: Yes. I need to update it. I couldn't find the code for TPPs that I should use. Can you send the link of the code where you have used them?

samples/deeplearning/conv1dopti_layer/Conv1dOpti-extension/Conv1dOpti.cpp
narendrachaudhary51: The matrix size is M = 64 (Tile size), N = 15 (Number of filters). I don't know how to check if the optimized path is being used for transpose.

samples/deeplearning/conv1dopti_layer/Conv1dOpti-extension/Bfloat16.h
alheinecke: intrinsics are here: 
https://github.com/hfp/libxsmm/blob/master/include/libxsmm_intrinsics_x86.h#L647
and usage here:
https://github.com/hfp/libxsmm/blob/master/src/template/libxsmm_dnn_bf16_macros_define.tpl.c#L16

However, I believe with TPP the whole file is not needed

narendrachaudhary51: Yes, you are right that I may not need this file at all.

samples/deeplearning/conv1dopti_layer/Conv1dOpti-extension/Conv1dOpti.cpp
narendrachaudhary51: Is it possible to lose some accuracy if I switch from scaler transform to TPP based transform (in FP32)?

alheinecke: only if it changes the order of your computation around them. they are just data movements and don't do any calculations on the the data.

samples/deeplearning/conv1dopti_layer/Conv1dOpti-extension/Bfloat16.h
narendrachaudhary51: Replaced with TPP kernel for conversion and removed the header file. Resolving conversation for now. Let me know if you have any additional comment.

samples/deeplearning/conv1dopti_layer/Conv1dOpti-extension/Conv1dOpti.cpp
narendrachaudhary51: I am using the TPP kernels now. So resolving this conversation too.

samples/deeplearning/conv1dopti_layer/Conv1dOpti-extension/Conv1dOpti.cpp
narendrachaudhary51: I use the TPP kernel for conversion now. Resolving this conversation.

samples/deeplearning/conv1dopti_layer/Conv1dOpti-extension/Conv1dOpti.cpp
narendrachaudhary51: I am using the libxsmm_dispatch_meltw_copy() kernel for initialization of zeros now. I tried using the 2D version but it threw an error or did not run correctly (maybe because I am trying to initialize a submatrix). So, I ran 1xN implementation in the loop for M time to initialize and MxN array. This gave me 30-40% performance loss.

samples/deeplearning/conv1dopti_layer/Conv1dOpti-extension/Conv1dOpti.cpp
narendrachaudhary51: Same issue with the initialization using libxsmm_dispatch_meltw_copy() kernel in a loop. 30-40% drop in performance.

samples/deeplearning/conv1dopti_layer/run.sh
narendrachaudhary51: The new commit has relative path. Resolving conversion now.

samples/deeplearning/conv1dopti_layer/Conv1dOpti-extension/Conv1dOpti.cpp
narendrachaudhary51: I could not find a TPP kernel that can change tensor layout/shape. Can you send me a link of a TPP kernel where you change the tensor layout/shape? I don't think I can use a transform kernel.

samples/deeplearning/conv1dopti_layer/Conv1dOpti-extension/Conv1dOpti.cpp
narendrachaudhary51: I replaced the scaler transform with mateltwise transform kernel. But, my weight updates (after backward pass) are wrong sometimes. It is weird because sometimes the results of weight update match with MKLDNN layer and sometimes half of them don't match. This was not the case with my scaler transform kernel. The weights always matched with MKLDNN layer.

samples/deeplearning/conv1dopti_layer/Conv1dOpti-extension/Conv1dOpti.cpp
egeor: Can you please let me know what is the error when you try the 2D approach? If you want to initialize a submatrix, you can do it by setting the proper LD.

samples/deeplearning/conv1dopti_layer/Conv1dOpti-extension/Conv1dOpti.cpp
egeor: Same as above, can you please point to the JIT error?

samples/deeplearning/conv1dopti_layer/Conv1dOpti-extension/Conv1dOpti.cpp
narendrachaudhary51: I have this configuration of Large Matrix and Submatrix.
Large Matrix - number of rows = 15, columns = 60000
Submatrix - number of rows = 15, columns = 64 (I need to transform this)
1st case (assuming row major layout) -  I set M = 15, N = 64 and ldo = 60000, ldi = NULL. This gives me a Seg fault error.
2nd case (assuming column major) - I set M = 64, N = 15 and ldo = 15, ldi=NULL. In this case the dispatch kernel returns a NULL pointer.

alheinecke: you need to set the ldi I would guess as you have an input and an output matrix.

narendrachaudhary51: But I am using this flag LIBXSMM_MELTW_FLAG_COPY_ZERO and not using any input matrix.  There is only an output matrix.

alheinecke: I mean you need at least ldo for 2d.
@egeor : do we have a standalone tester for this?

egeor: I believe the problem was some LD check, I fixed it in master. Please pull/rebuild; you should specify only ldo for the "zero" kernel. For future reference, please use LIBXSMM_VERBOSE=-1, this would reveal the error right away since the proper error message would be shown...

samples/deeplearning/conv1dopti_layer/Conv1dOpti-extension/Conv1dOpti.cpp
narendrachaudhary51: I get wrong results with the VNNI Transform kernel. The parameters in use. I use LIBXSMM_MELTW_FLAG_TRANSFORM_NORM_TO_VNNI flag.
M = 464, N = 16, ldi = 60400, ldo = 512
Is this expected?

alheinecke: Hmm,,, I used the harness (extended it to arbitrary ldi/ldo) in https://github.com/hfp/libxsmm/tree/master/samples/eltwise
./eltwise_transform V 2 464 16 60400 512
this one is giving correct results and also the assembly looks fine:
XDIS 0: DATAXFER  BASE       4C8B07                   mov r8, qword ptr [rdi]
XDIS 3: DATAXFER  BASE       4C8B4F08                 mov r9, qword ptr [rdi+0x8]
XDIS 7: DATAXFER  BASE       49C7C201000000           mov r10, 0x1
XDIS e: KMASK     AVX512VEX  C4C17B92CA               kmovd k1, r10d
XDIS 13: DATAXFER  BASE       49C7C201000000           mov r10, 0x1
XDIS 1a: KMASK     AVX512VEX  C4C17B92D2               kmovd k2, r10d
XDIS 1f: DATAXFER  BASE       48C7C200000000           mov rdx, 0x0
XDIS 26: BINARY    BASE       4883C202                 add rdx, 0x2
XDIS 2a: DATAXFER  BASE       48C7C000000000           mov rax, 0x0
XDIS 31: BINARY    BASE       4883C001                 add rax, 0x1
XDIS 35: DATAXFER  AVX512EVEX 62D1FF896F00             vmovdqu16 xmm0{k1}{z}, xmmword ptr [r8]
XDIS 3b: DATAXFER  AVX512EVEX 62D1FF896F88E0D70100     vmovdqu16 xmm1{k1}{z}, xmmword ptr [r8+0x1d7e0]
XDIS 45: DATAXFER  AVX512EVEX 62D1FF0A7F01             vmovdqu16 xmmword ptr [r9]{k2}, xmm0
XDIS 4b: DATAXFER  AVX512EVEX 62D1FF0A7F8902000000     vmovdqu16 xmmword ptr [r9+0x2]{k2}, xmm1
XDIS 55: BINARY    BASE       4983C002                 add r8, 0x2
XDIS 59: BINARY    BASE       4983C104                 add r9, 0x4
XDIS 5d: BINARY    BASE       483DD0010000             cmp rax, 0x1d0
XDIS 63: COND_BR   BASE       7CCC                     jl 0x31
XDIS 65: BINARY    BASE       4981C020AC0300           add r8, 0x3ac20
XDIS 6c: BINARY    BASE       4981C1C0000000           add r9, 0xc0
XDIS 73: BINARY    BASE       4883FA10                 cmp rdx, 0x10
XDIS 77: COND_BR   BASE       7CAD                     jl 0x26
XDIS 79: RET       BASE       C3                       ret 

0x1d0: 464
0x1d7e0: 120800

narendrachaudhary51: It is working after the race condition fix in my code. Changes added to latest commit. Resolving the conversation now.

samples/deeplearning/conv1dopti_layer/Conv1dOpti-extension/Conv1dOpti.cpp
narendrachaudhary51: It is working now. Changes added to latest commit. Resolving the conversation now.

samples/deeplearning/conv1dopti_layer/Conv1dOpti-extension/Conv1dOpti.cpp
narendrachaudhary51: It is working now. No significant performance drop. Changes added to latest commit. Resolving the conversation now.

samples/deeplearning/conv1dopti_layer/Conv1dOpti-extension/Conv1dOpti.cpp
narendrachaudhary51: It is working after the race condition fix in my code. Changes added to latest commit. Resolving the conversation now.

alheinecke (commented):
Do I see it correctly that we need to close on the following items?: 
a) ReLU TPP still needs to be integrated
b) Padding TPP needs to be defined
c) identify which TPP with what problemsizes do negatively impact performance
d) getting rid of "USE_TPP" define

samples/deeplearning/conv1dopti_layer/Conv1dOpti-extension/Conv1dOpti.cpp
narendrachaudhary51: I used the transpose TPP and it is working now. I was able to use transpose TPP in place of all permute kernels. Changes present in the latest commit. Resolving conversation now.

samples/deeplearning/conv1dopti_layer/Conv1dOpti-extension/Conv1dOpti.cpp
narendrachaudhary51: I was able to use the transpose TPP kernels to permute dimensions. However, the weight kernel flip/reverse part required two loops. It is not clean as we are just increasing the amount of operations. Check out line 324-386 in Conv1dOpti.cpp file. 
Do we have a reverse copy kernel?   Something that does the following without a loop.
a1, a2, a3, a4, a5   -------------> a5, a4, a3, a2, a1

narendrachaudhary51:
@alheinecke, I changed the permute loops into transform TPPs. 
a)  I am facing some error in relu due to the mask pointer type. I will apply it correctly and integrate it after correction.
b) Yes. I will define the virtual padding TPP. 
c) Currently, I do not see any performance drop. I made 20 iteration runs with and without TPP. The performance difference was 0.5-1% (or less) on cascade lake in FP32 run. So performance difference with TPP kernels seems negligible (atleast for the parameters that I use).     
d)  I will also remove USE_TPP after relu integration and checking everything.

alheinecke:
Was the ReLU is resolved, do you need @egeor 's help?

narendrachaudhary51:
I just added the relu kernel. I am currently using the same TPP for both backward and forward pass (with backward flag). It is working for now. I have mentioned the error with FWD flag to @egeor.

samples/deeplearning/conv1dopti_layer/Conv1dOpti-extension/Conv1dOpti.cpp
narendrachaudhary51: Replaced the virtual padding loops with TPP kernels. Resolving the issue now.

samples/deeplearning/conv1dopti_layer/Conv1dOpti-extension/Conv1dOpti.cpp
narendrachaudhary51: Resolving the issue for now.

narendrachaudhary51:
I have replaced all the parts of existing code with TPP kernels. Now, I need to check the performance and remove USE_TPP define after that.

alheinecke:
great, thanks. Please let us know when done and then we can merge.

narendrachaudhary51:
I made all the changes and removed the older non-TPP code. Code is ready for integration from my side.

alheinecke (approved):

PR #431, opened 2020-11-25
added Zipf distribution and LLC performance counters

ibhati:


samples/deeplearning/sparse_adagrad_fused/sparse_adagrad.cpp
```diff
@@ -273,7 +274,53 @@ void sparse_transpose(EmbeddingInOut *eio)
...
+      k = k + (1.0 / pow((double) i, alpha));
+    k = 1.0 / k;
+
+    sum_probs = (double *) my_malloc((M+1)*sizeof(double), alignment);
+    sum_probs[0] = 0;
+    for (i=1; i<=M; i++) {
+      sum_probs[i] = sum_probs[i-1] + k / pow((double) i, alpha);
+    }
+    init_done = 1;
+  }
```
ddkalamk, line 299: Guess, this assumes M won't change across zipf_dist call. Though for the current setup it is true, it can cause an error if we extend main() to allow variable size tables. Probably it is better to compute it in every call for now. or leave a comment somewhere..

ibhati:
Yes, good point. Updated code to recalculate distribution whenever M changes.

samples/deeplearning/sparse_adagrad_fused/sparse_adagrad.cpp
ibhati: Yes, good point. Updated code to recalculate distribution whenever M changes.

PR #436, opened 2020-12-02
Improve 16-bit flat->flat transpose sequence

adelmanm:
Improve 16-bit flat->flat transpose sequence
Add LIBXSMM_X86_INSTR_VBROADCASTI64X2 support


src/generator_mateltwise_transform_avx_avx512.c
```diff
@@ -579,111 +621,96 @@ void libxsmm_generator_transform_norm_to_normt_16bit_avx512_microkernel( libxsmm
   unsigned long long l_mask = 0;
 
   /* optimized shuffle network for SIMD aligned sizes */
-  if ( (i_mateltwise_desc->m % 32 == 0) && (i_mateltwise_desc->n % 16 == 0) ) {
```
alheinecke, line 582: Can you please leave the old code under an #ifdef in case we need this with masking for the more general case (e.g. not a full SIMD Lane) in future. Or will this all be possible with quarter loads?

alheinecke (commented):
Thanks for your contribution.

Please see my comment on if we should keep the old code for now under #ifdef as we might need it in one the masked variants in future?

src/generator_mateltwise_transform_avx_avx512.c
adelmanm: We can still do masking on quarter loads at dword granularity. Even if we need word granularity, we can still do it with masked load followed by broadcast. The old code can still be found in the git history if needed.

Having said all that, if you still want the old code present we can put it back under #ifdef.

PR #438, opened 2020-12-10
Add optimized VNNI->VNNI^T code

adelmanm:


src/generator_mateltwise_transform_avx_avx512.c
```diff
@@ -190,6 +190,55 @@ void libxsmm_generator_transform_08way_permutevar_network_avx512( libxsmm_genera
   }
 }
 
+LIBXSMM_API_INTERN
+void libxsmm_generator_transform_Xway_permute_network_avx512( libxsmm_generated_code* io_generated_code,
```
alheinecke, line 194: The code has already "libxsmm_generator_transform_16way_permute_network_avx512", can you please make libxsmm_generator_transform_16way_permute_network_avx512 use libxsmm_generator_transform_Xway_permute_network_avx512 or integrated it directly where libxsmm_generator_transform_16way_permute_network_avx512 is used. I like this more general libxsmm_generator_transform_Xway_permute_network_avx512 and think we should replace the limited libxsmm_generator_transform_16way_permute_network_avx512 is used

src/generator_mateltwise_transform_avx_avx512.c
```diff
@@ -319,6 +368,38 @@ void libxsmm_generator_transform_Xway_quarter_load_blend_avx512( libxsmm_generat
   }
 }
 
+LIBXSMM_API_INTERN
+void libxsmm_generator_transform_Xway_half_load_blend_v2_avx512( libxsmm_generated_code* io_generated_code,
```
alheinecke, line 372: There is already a libxsmm_generator_transform_Xway_half_load_blend_avx512 and this one seems substantially different is there a chance to write one libxsmm_generator_transform_Xway_half_load_blend_avx512 which serves both purposes, e.g. when providing the register pairs as we do for shuffle/permutes? Ideally we would have only one general libxsmm_generator_transform_Xway_half_load_blend_avx512.

alheinecke (commented):
Looks very nice! Thanks.

I was just wondering if we can do a little bit extra to avoid code duplication by generalizing the two shuffle networks in question.

src/generator_mateltwise_transform_avx_avx512.c
zbarukh: Right. The main difference between 2 applications of this routine is that the original one loads and blends i_ways cache lines, while the new one - 2 * i_ways half cache lines. What is similar in these versions is that the offset between the "load" and "blend" halves is constant.

I agree with your idea to make the unified version similar to shuffle/permute networks. I suggest the following: Add 2 arguments to the interface: load_disp[i_ways] and blend_disp. Then, for every register set:

load displacement = i_ld * load_disp[i / 2] + 32 * load_disp[i % 2]
blend displacement = load displacement + blend_disp * i_ld

Also, no extra registers will be used.

This will allow using the same interface for both applications. If you are fine with this, I will implement it.

alheinecke: Yes, this sounds great, please go ahead and I believe we can even use it then in a revised version of the 32bit transposes.

zbarukh: Done. PR will be ready by tomorrow.

src/generator_mateltwise_transform_avx_avx512.c
zbarukh: Done. libxsmm_generator_transform_16way_permute_network_avx512 is absorbed into more general libxsmm_generator_transform_Xway_permute_network_avx512. The latter supports any even number of i_ways up to 20. It uses i_ways/2 additional registers.

I hope we will prepare another PR by tomorrow.

src/generator_mateltwise_transform_avx_avx512.c
adelmanm: Pushed Barukh's changes

alheinecke (approved):

PR #442, opened 2020-12-15
Add optimized 32-bit transpose sequence

adelmanm:


alheinecke (approved):
Awesome thanks for the contribution

PR #443, opened 2020-12-21
Add support for 16-bit flat->flat transpose with arbitrary dimensions

adelmanm:


alheinecke (approved):
Great, thank you and happy holidays!

PR #444, opened 2020-12-22
Add NUMA aware "first touch" initialization in f32 MLP version

nshustrov:


alheinecke (approved):
Thanks for you contribution.

PR #445, opened 2020-12-23
Add arbitrary dimensions support for transpsoe sequences

adelmanm:
Add arbitrary dimensions support for 16-bit vnni->vnni^T sequence
Add arbitrary dimensions support for 32-bit transpose
Fix functional correctness bug in 32-bit transpose


alheinecke (approved):
Great! Many thanks for the contribution.

PR #452, opened 2021-01-09
mlp_example_bf16_avx512_fused_trans_numa: make parallelization strategy be set by env. variable

adelmanm:
…gy be set by env. variable


alheinecke (commented):
can you please add some example or comment how to the env array should be structured and some default values to run the code. something along the lines on a 2s 28 core machine this is the best setting, etc.

zbarukh:
Sure, will add shortly.

zbarukh:
Question about default values. Do you want them to be part of the code? In which case we need to get back all the code which was before, as these values depend on #threads.

adelmanm:
I think it is better to have them in comment, e.g for the different values that were in the previous code (which were achieved after some tuning), document in comment what the values would be for the new command line. It can then be improved based on autotuning results.

zbarukh:
Done.

alheinecke (approved):

zbarukh:
I would like to extend this feature for other mlp_tpp examples (currently, it is supported only in mlp_example_bf16_avx512_fused_trans_numa). If there is no objection, I need you suggestion where to put a common function which will be called by all examples.

alheinecke:
As long as all the current work distribution for specific core counts remain in the code, that's fine, e.g. a special if branch in the end which attempt to read the env variable and then sets the values accordingly.

PR #459, opened 2021-02-06
Improve the unique value detection in the sparse register generator.

FreddieWitherden:
Currently we consider +x and -x to be two distinct unique values.  This doubles the number of required registers in cases where the unique values are symmetric (as often happens to be the case).

With this PR we only store the absolute value and perform the negation as required by issuing a different FMA instruction.

The number of registers required for several key classes of matrices is now below 15 and so we might want to consider also opening up the sparse path on AVX2 capable systems.


alheinecke (approved):
Thanks for your contribution!

PR #473, opened 2021-04-12
Dynamically decide between sparse and dense kernels in dfsspmdm.

FreddieWitherden:
The dfsspmdm now attempts to JIT both sparse and dense kernels and, if both are available, it benchmarks them to decide which to proceed with.  The code is functional, although I would like to get feedback before writing the single precision variant.

One area that needs work is how we release/free the slower kernel which we do not need.  Specifically, how to be sure that we do not deallocate a kernel which someone else still has a reference to.  As such the deallocation code is currently commented.

### DIFF
--- a/src/libxsmm_fsspmdm.c
+++ b/src/libxsmm_fsspmdm.c
@@ -13,6 +13,38 @@
 #include "libxsmm_main.h"
 
 
+/* Double precision AVX-512 lane broadcasts */
+LIBXSMM_APIVAR_DEFINE(const unsigned int dperm[]) = {
+  0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1,
+  2, 3, 2, 3, 2, 3, 2, 3, 2, 3, 2, 3, 2, 3, 2, 3,
+  4, 5, 4, 5, 4, 5, 4, 5, 4, 5, 4, 5, 4, 5, 4, 5,
+  6, 7, 6, 7, 6, 7, 6, 7, 6, 7, 6, 7, 6, 7, 6, 7,
+  8, 9, 8, 9, 8, 9, 8, 9, 8, 9, 8, 9, 8, 9, 8, 9,
+  10, 11, 10, 11, 10, 11, 10, 11, 10, 11, 10, 11, 10, 11, 10, 11,
+  12, 13, 12, 13, 12, 13, 12, 13, 12, 13, 12, 13, 12, 13, 12, 13,
+  14, 15, 14, 15, 14, 15, 14, 15, 14, 15, 14, 15, 14, 15, 14, 15
+};
+
+/* Single precision AVX-512 lane broadcasts */
+LIBXSMM_APIVAR_DEFINE(const unsigned int sperm[]) = {
+  0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+  1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,
+  2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2,
+  3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3,
+  4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4,
+  5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5,
+  6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6,
+  7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7,
+  8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8,
+  9, 9, 9, 9, 9, 9, 9, 9, 9, 9, 9, 9, 9, 9, 9, 9,
+  10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10,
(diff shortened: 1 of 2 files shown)


src/libxsmm_fsspmdm.c
```diff
@@ -51,86 +68,159 @@ LIBXSMM_API libxsmm_dfsspmdm* libxsmm_dfsspmdm_create(
...
-  if (new_handle->kernel != 0) {
-    /* allocate 8 * 512-bit permute operands if not stored in registers */
-    new_handle->permute_operands = (unsigned int*)libxsmm_aligned_malloc(8*16*sizeof(unsigned int), 64);
+  /* Free CSR */
+  free( a_csr_values );
+  free( a_csr_rowptr );
+  free( a_csr_colidx );
+
+  if ( 0 != k_sparse ) {
     /* store permute operands */
```
FreddieWitherden, line 121: I'd like to statically allocate this array, but we'd need to make sure that the compiler aligns it on a 64-byte boundary.

alheinecke: We can do stack allocation (if this is general) not an issue: __attribute__((aligned(64)))

src/libxsmm_fsspmdm.c
```diff
@@ -51,86 +68,159 @@ LIBXSMM_API libxsmm_dfsspmdm* libxsmm_dfsspmdm_create(
...
+  }
+  /* Dense only */
+  else if ( 0 != k_dense && 0 == k_sparse ) {
+    new_handle->N_chunksize = N_dense;
+    new_handle->kernel = k_dense;
+    new_handle->a_dense = aa_dense;
+    libxsmm_free( permop );
+  }
+  /* Both; benchmark */
+  else if ( 0 != k_dense && 0 != k_sparse ) {
```
alheinecke, line 160: here we could add some parameter to the create call if we want to actually do this runtime tuning.

alheinecke (commented):
that looks good to me and I don't think we want to free the kernels, at least not the dense one as this one is cached and maybe useful for other calls in the application. The sparse kernels are not put into the code registry https://github.com/hfp/libxsmm/blob/master/src/libxsmm_main.c#L5091 so we can free it, if we decide to use the dense kernel.

FreddieWitherden:
Updated.  Permute operands are now static arrays and single precision is supported.

src/libxsmm_fsspmdm.c
FreddieWitherden: I thought about that, but am wary of breaking ABI compatibility.

alheinecke: we could add a 2nd call for now with the additional flag. We are right now cleaning up meaning things in the code as we prepare to full release of tensor processing primitives (TPP), https://arxiv.org/abs/2104.05755  with LIBXSMM v2.0 withing the next 2-3 months. This will allow for ABI breaks.

alheinecke:
I'm fine with the code merging it. Just waiting on your comment above on the ABI discussion.

FreddieWitherden:
Lets leave the ABI stable for the moment; as it is possible that we'll want to change it again when ARM support is added on the sparse path so it would be nice to break all at once.

src/libxsmm_fsspmdm.c
```diff
@@ -1,18 +1,50 @@
...
 /* Alexander Heinecke (Intel Corp.)
-******************************************************************************/
+ ******************************************************************************/
 #include "generator_spgemm_csr_asparse_reg.h"
 #include <libxsmm_fsspmdm.h>
 #include "libxsmm_main.h"
 
 
+/* Double precision AVX-512 lane broadcasts */
+static const unsigned int dperm[] __attribute__((aligned (64))) = {
```
hfp, line 17: Perhaps use the following:

```c
LIBXSMM_ALIGNED(const unsigned int dperm[], 64) = {
```

Even better, use `LIBXSMM_APIVAR_DEFINE` like:

```
LIBXSMM_APIVAR_DEFINE(const unsigned int dperm[]) = {
```

The latter is cacheline-aligned aka 64 Byte (we do not have this kind of macro with explicit alignment).

We generally do not use raw `__attribute__` syntax because of portability, the `static` keyword at file-level scope is not necessary/superfluous, and regarding `LIBXSMM_APIVAR_DEFINE`, there is some need to wrap-up "global" variables (even at file scope) because LIBXSMM supports header-only for C and C++. The latter needs special handling of variables different between platforms/compilers, C, and C++.

src/libxsmm_fsspmdm.c
```diff
@@ -51,86 +99,135 @@ LIBXSMM_API libxsmm_dfsspmdm* libxsmm_dfsspmdm_create(
...
+    int use_sparse = 1;
+    libxsmm_timer_tickint t1, t2, t3;
+    double* B = (double*)libxsmm_aligned_malloc((size_t)K * (size_t)ldb * sizeof(double), 64);
+    double* C = (double*)libxsmm_aligned_malloc((size_t)M * (size_t)ldc * sizeof(double), 64);
+    void* fp;
 
-    if (0 != xgemm_desc) {
-      new_handle->kernel = libxsmm_create_dcsr_reg(xgemm_desc, a_csr_rowptr, a_csr_colidx, a_csr_values);
+    if (0 != B && 0 != C ) {
+      for ( i = 0; i < K; i++ )
```
hfp, line 180: We would like to ban multi-line constructs without block statement/curly braces. It's particularly an issue with if-statements where indentation can suggest some flow that is not in effect. Though, we allow short one-liners of such kind.

src/libxsmm_fsspmdm.c
hfp: This patch does not seem to change the ABI and also does not add to the signatures of public functions. Also, if we know the deployed applications and the functions used by such applicatoins, I would not be too worried about ABI changes if it only touches more rarely used functionality.

FreddieWitherden:
Updated.

src/libxsmm_fsspmdm.c
hfp: Perhaps the nested `for` loops would be nicer with curly braces... don't worry too much, I can take it when the PR is merged.

@alheinecke LGTM

PR #476, opened 2021-04-16
A64fx

breuera:


src/generator_aarch64_instructions.c
```diff
@@ -646,8 +651,18 @@ void libxsmm_aarch64_instruction_sve_pcompute( libxsmm_generated_code*
     code[code_head]  = (unsigned int)(0xffffff00 & i_pred_instr);
     /* setting Rd */
     code[code_head] |= (unsigned int)(0xf & i_pred_reg);
-    /* setting pattern */
-    code[code_head] |= (unsigned int)((0x1f & i_pattern) << 5);
+    if( i_pred_instr == LIBXSMM_AARCH64_INSTR_SVE_PTRUE ) {
```
alheinecke, line 654: Can we avoid adding ifs on the entire 32bit instruction define. In this case, would it be sufficient to check the numbers registers?

alheinecke (commented):
I left one comment to see if we can refactor one if.

Does it work for F64 and F32 (it seems so)?

breuera:
> I left one comment to see if we can refactor one if.

Replaced the ifs on the instruction by checks w.r.t. the number of registers in the instruction.
Didn't check if this is generic enough for future instructions in pcompute but we can adjust later if needed.

> Does it work for F64 and F32 (it seems so)?

I am assuming you are asking about the matrix-kernels not the instructions.
Yes, they work in FP32 and FP64.
The standard ``whilelt`` examples, e.g., slide 16 @ https://gitlab.com/arm-hpc/training/arm-sve-tools/-/blob/master/Slides/5%20-%20Advanced%20SVE.pptx, use the instruction slightly different though.
For example, if activating the first three lanes in SP they'd use ``pX.s`` in``whilelt``, giving:
```0000[...]0000100010001```
DP similar:
```00000000[...]0000000010000000100000001```
I am just setting all bits using ``pX.b``, e.g., for FP32:
```0000[...]0111111111111```
The additional bits are ignored.

Don't think that this make a difference in performance but haven't checked either.

alheinecke:
create and thanks, cleared to merge

PR #479, opened 2021-05-04
added zipf indices distribution. 

ibhati:
Added ZIPF distribution on indices
Fixed number of generated indices (random range up to 2*P)


alheinecke:
Thanks for your contribution

PR #482, opened 2021-05-17
Groupnorm Code

narendrachaudhary51:
I have added the groupnorm code in samples/equation/equation_groupnorm.c file.
It have followed a design that is similar to equation_layernorm.c file. 
Groupnorm code is working for FP32 and BF16 precision according to my testing. You can also test the code by using a command line. I have not removed some of the layernorm code because I use it to test BF16 implementation.


alheinecke (changes requested):
I don't think it makes sense to touch the conv1dopti in this PR. Why is that needed?

narendrachaudhary51:
@alheinecke I was using the using the older branch and pushed changes from it. I only corrected a one line change in conv1dopti layer example code (making ms time unit to sec). Do you want to revert that change and resubmit it separately? Or do you want me to create a separate branch for Groupnorm?

alheinecke:
OK, then it's fine if you are saying that this is bugfix. Ideally, we should just do 2 PRs, but if you are fine with one, I will just merge.

narendrachaudhary51:
@alheinecke I fine with it for now. Next time, I will create a different PR.

PR #486, opened 2021-05-21
Support double-wide chunk sizes in spgemm_csr_asparse_reg.

FreddieWitherden:
This enables chunk sizes larger than the natural vector width to be passed to `libxsmm_generator_spgemm_csr_asparse_reg`.  This may enable better performance for certain multiplications since there can now be two explicit accumulation chains (whereas previously we depended on the OOE engine to find independent chains for us).

Note that currently none of the higher level routines request these larger chunks.  This will be the subject of a future PR.


alheinecke (approved):

alheinecke:
Thanks for your contribution.

PR #498, opened 2021-07-26
Blocked  and Parallel Batch normalization code. 

narendrachaudhary51:


narendrachaudhary51:
@alheinecke The code is done for now. You can review and incorporate it. I tried 4*CP for middle loop but it didn't give me a performance advantage. However, reducing HW block size from 49 to 28 gives some improvement.

alheinecke (approved):

PR #500, opened 2021-07-30
AVX512vl blocking fix

deeptiag1:
SP and Int8 for AVX512VL


deeptiag1:
Rebased it to avx512vl

PR #517, opened 2021-09-24
Disable prefetching for sparse register kernels

FreddieWitherden:
This disables pre-fetching for the sparse register kernels which results in a slight performance improvement and appreciably reduced code size (the primary benefit) on modern AVX-512 architectures.

### DIFF
--- a/src/generator_spgemm_csr_asparse_reg.c
+++ b/src/generator_spgemm_csr_asparse_reg.c
@@ -131,7 +131,12 @@ void libxsmm_generator_spgemm_csr_asparse_reg( libxsmm_generated_code*         i
     l_breg_unique = 32 - l_n_blocking;
     l_base_acc_reg = 32 - l_n_blocking;
     l_bcast_reg = l_base_acc_reg - 1;
-    l_prefetch = 0;
+    if ( (io_generated_code->arch == LIBXSMM_X86_AVX512_KNL) ||
+         (io_generated_code->arch == LIBXSMM_X86_AVX512_KNM)    ) {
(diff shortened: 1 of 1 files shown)


src/generator_spgemm_csr_asparse_reg.c
```diff
@@ -131,7 +131,7 @@ void libxsmm_generator_spgemm_csr_asparse_reg( libxsmm_generated_code*         i
     l_breg_unique = 32 - l_n_blocking;
     l_base_acc_reg = 32 - l_n_blocking;
     l_bcast_reg = l_base_acc_reg - 1;
-    l_prefetch = 1;
+    l_prefetch = 0;
```
alheinecke, line 134: I guess this was measured on SKX/CLX/ICX ? The code already was written for KNL/KNM. I'm happy with merging in that way. However, we simple if on != KNL and != KNM we can get the clean solution.

FreddieWitherden: Yes, benchmarked on SKX.  Sadly I do not have access to any KNL/KNM hardware as we retired them a few years ago now.

alheinecke: OK then i suggest the following code:

```suggestion
    if ( (io_generated_code->arch == LIBXSMM_X86_AVX512_KNL) ||
         (io_generated_code->arch == LIBXSMM_X86_AVX512_KNM)    ) {
      l_prefetch = 1;
    } else {
      l_prefetch = 0;
    }
```

PR #522, opened 2021-10-05
Unary Exp and Gelu Framework : Part1

deeptiag1:
Code styling fixes with test scripts in Part2


src/generator_aarch64_instructions.c
```diff
@@ -1308,6 +1310,25 @@ void libxsmm_aarch64_instruction_cond_jump_back_to_label( libxsmm_generated_code
   }
 }
 
+LIBXSMM_API_INTERN
```
alheinecke, line 1313: This function is not load a vector of constants. It's broadcasting a constant scalar. I like that function, however, relu and dropout need such a thing as well so it would be great to back port there as well. At least it should be renamed to what it's doing.
I'm not convinced this is the right semantics what we need for some of the usages (where we really need to load a vector)... e.g. in GeLU

src/generator_common_aarch64.c
```diff
@@ -644,6 +644,572 @@ void libxsmm_generator_xoshiro128p_f32_aarch64_asimd( libxsmm_generated_code* io
...
+  libxsmm_aarch64_instruction_full_vec_load_of_constants ( io_generated_code, i_vec_c1, i_gp_reg_tmp, i_tupletype , vec_c1  );
+  libxsmm_aarch64_instruction_full_vec_load_of_constants ( io_generated_code, i_vec_c2, i_gp_reg_tmp, i_tupletype , vec_c2 );
+  libxsmm_aarch64_instruction_full_vec_load_of_constants ( io_generated_code, i_vec_c3, i_gp_reg_tmp, i_tupletype , vec_c3 );
+  libxsmm_aarch64_instruction_full_vec_load_of_constants ( io_generated_code, i_vec_expmask, i_gp_reg_tmp, i_tupletype , vec_expmask );
+  libxsmm_aarch64_instruction_full_vec_load_of_constants ( io_generated_code, i_vec_hi_bound, i_gp_reg_tmp, i_tupletype , vec_hi_bound );
+  libxsmm_aarch64_instruction_full_vec_load_of_constants ( io_generated_code, i_vec_lo_bound, i_gp_reg_tmp, i_tupletype , vec_lo_bound );
+}
+
+LIBXSMM_API_INTERN
+void libxsmm_generator_exp_ps_3dts_aarch64( libxsmm_generated_code*                        io_generated_code,
```
alheinecke, line 1052: In this routine: can we clean-up the not needed comments.

alheinecke (changes requested):
some minor comments inside

alheinecke:
BTW, CVT instruction still didn't carry the full instruction name which should happen

alheinecke (changes requested):
This seems to be broken as it removes the file. Can you please remove this commit from the PR?

PR #523, opened 2021-10-05
Unary and Gelu Commit: Part1

deeptiag1:
Unary Exp is fully validated


deeptiag1:
@alheinecke: Ported Comments from Part522: This function is not load a vector of constants. It's broadcasting a constant scalar. I like that function, however, relu and dropout need such a thing as well so it would be great to back port there as well. At least it should be renamed to what it's doing.
I'm not convinced this is the right semantics what we need for some of the usages (where we really need to load a vector)... e.g. in GeLU

alheinecke:
do you want me to do review/merge or should I wait until the load broadcast routine is moved ?

deeptiag1:
Fixing the load broadcast routine. Checking the requirements in Gelu and Relu to sync in one shot

alheinecke:
do you have some updates on the load broadcast routine (moving to the correct file, and renaming?)

PR #552, opened 2021-11-08
Rework asparse_reg.

FreddieWitherden:
This PR represents a substantial rework of asparse_reg.  It now includes support for m blocking by grouping together rows of A.  This opens up more opportunities for ILP.  Further, a range of measures are introduced to reduce the size of generated code and improve performance.


FreddieWitherden:
The code now includes the functionality of #525.  Still need to perform some testing on AVX2 class systems.

Todo on the x86 side:
 - Further instruction compression; currently our broadcasts can be 10 bytes on a bad day which given we're often outside of the op-cache isn't great.  Using GPR's I should be able to get this down.
 - A broadcast elision when B is resident in a register since we can get the FMA instruction to do it for us.
 - Factoring out of the n-blocking argument so the kernel can loop over N itself.
 - Better grouping/register allocation logic.

There is quite a bit more to do on the ARM side (performance is sometimes worse than GiMMiK on an N1 core), but that should not stop us from merging.

FreddieWitherden:
Should be ready for review.

alheinecke (commented):
I think we still need the read-only protection the other branch to merge this back, correct?

alheinecke:
I have rebased this PR to https://github.com/hfp/libxsmm/tree/const_data_in_kernel and I will merge and then we can work on fixing all the issue creeping up from static analysis etc. Sounds good?

FreddieWitherden:
> I have rebased this PR to https://github.com/hfp/libxsmm/tree/const_data_in_kernel and I will merge and then we can work on fixing all the issue creeping up from static analysis etc. Sounds good?

Sounds good.  The static analyzer is going to hate this one!

PR #556, opened 2021-11-15
First SVE kernels

AntonioNoack:
Hello,

I'm currently working on SVE support for the TPP kernels. I've heard from Alex Breuer, that there will be a second team working on SVE, so I shouldn't wait with pushing my progress until all kernels are working.

I've implemented set-zero, copy, increment, square, 1/x and sqrt currently.
This pull request also adds a few SVE instructions for these, which haven't been part of LIBXSMM yet.

As a testing suite for performance and correctness, I added a sample/sve directory, but Mr. Breuer told me, that it may be possible to integrate the tests into the samples/eltwise directory in the future.

Known/found issues:
- the A64FX architecture is not automatically detected
- only the unary SVE kernels, that I've implemented, are currently working with the A64FX architecture
- the 1/x function for ASIMD seems to be only an approximation
- when switching architectures, the kernel cache is not cleared automatically

Thank you,
Antonio Noack


samples/sve/sve.cpp
```diff
@@ -0,0 +1,214 @@
+/******************************************************************************
```
alheinecke, line 1: Test code should be all C89 or C99 as the whole projection is C89. 
What is the reason for not using the testers incl. the test scripts in samples/eltwise ? These are arch agnostic and work for ARM and X86.

FreddieWitherden:
Some of the instructions overlap with the PR of #552 .  I am also led to believe that several of the added intrinsics have extremely poor performance on A64FX (sqrt in particular).

src/generator_aarch64_instructions.c
```diff
@@ -741,9 +741,53 @@ void libxsmm_aarch64_instruction_sve_compute( libxsmm_generated_code*        io_
...
+  unsigned char l_is_predicated = false;
+  unsigned char l_is_type_specific = true; /* the only exception currently is xor */
   switch ( i_vec_instr ) {
+    /* usual, binary functions, unpredicated */
     case LIBXSMM_AARCH64_INSTR_SVE_FMLA_V:
+    case LIBXSMM_AARCH64_INSTR_SVE_FMUL_V:
+    case LIBXSMM_AARCH64_INSTR_SVE_FRECPS_V:
+      break;
     case LIBXSMM_AARCH64_INSTR_SVE_EOR_V:
+      l_is_type_specific = false;
```
alheinecke, line 758: We should not case on the instructions in the main part of the encoder as this slows down the encoder. There is right now another PR open as SVE used case on instructions name before. 

Please look at ASIMD or AVX512F as example on how we can follow the general encoding scheme of the machine to generate a

src/generator_aarch64_instructions.c
```diff
@@ -752,23 +796,28 @@ void libxsmm_aarch64_instruction_sve_compute( libxsmm_generated_code*        io_
 
   if ( io_generated_code->code_type > 1 ) {
     unsigned int code_head = io_generated_code->code_size/4;
-    unsigned int* code     = (unsigned int *)io_generated_code->generated_code;
+    unsigned int* code     = (unsigned int *) io_generated_code->generated_code;
 
     /* fix bits */
-    code[code_head]  = (unsigned int)(0xffffff00 & i_vec_instr);
```
alheinecke, line 758: Please use code[code_head] through out the routine such that we have alignment with all other encoders

src/generator_aarch64_instructions.h
```diff
@@ -344,8 +344,15 @@
 #define LIBXSMM_AARCH64_INSTR_SVE_LD1RW_I_OFF    0x8540c086
 #define LIBXSMM_AARCH64_INSTR_SVE_LD1RD_I_OFF    0x85c0e086
 /* define SVE compute instructions */
-#define LIBXSMM_AARCH64_INSTR_SVE_FMLA_V         0x65200083
-#define LIBXSMM_AARCH64_INSTR_SVE_EOR_V          0x04a03003
+#define LIBXSMM_AARCH64_INSTR_SVE_FMLA_V         0x65200083 /* fused multiply-add */
```
alheinecke, line 347: Here is overlap with #552, I guess as #552 is close to got to master, we merge #552 into master and then merge master into aarch64_eltwise

src/generator_mateltwise.c
```diff
@@ -12,15 +12,19 @@
...
 #include "generator_mateltwise_sse_avx_avx512.h"
 #include "generator_mateltwise_aarch64.h"
+#include "generator_mateltwise_aarch64_sve.h"
 
 LIBXSMM_API
 void libxsmm_generator_mateltwise_kernel( libxsmm_generated_code*          io_generated_code,
                                           const libxsmm_meltw_descriptor*  i_mateltw_desc ) {
+  unsigned int l_arch = io_generated_code->arch;
   /* generate kernel */
-  if ( (io_generated_code->arch >= LIBXSMM_X86_GENERIC) && (io_generated_code->arch <= LIBXSMM_X86_ALLFEAT) ) {
```
alheinecke, line 20: please keep this as this aligns with the coding style in the other generators

src/generator_mateltwise_aarch64.c
```diff
@@ -20,67 +20,70 @@
 #include "libxsmm_main.h"
 
 LIBXSMM_API_INTERN
-void libxsmm_generator_mateltwise_aarch64_update_micro_kernel_config_vectorlength( libxsmm_generated_code*           io_generated_code,
+unsigned int libxsmm_generator_mateltwise_aarch64_get_type_size(libxsmm_generated_code* io_generated_code, unsigned char type){
```
alheinecke, line 23: Please use: https://github.com/hfp/libxsmm/blob/master/src/libxsmm_main.c#L1689 and remove this routine

src/generator_mateltwise_aarch64.c
```diff
@@ -20,67 +20,70 @@
...
+      return 1;
+    default:
+      /* what if io_generated_code is null? */
+      LIBXSMM_HANDLE_ERROR( io_generated_code, LIBXSMM_ERR_UNSUP_DATATYPE );
+      return 0;
+  }
+}
+
+LIBXSMM_API_INTERN
+unsigned int libxsmm_generator_mateltwise_aarch64_sve_get_vlen(){
```
alheinecke, line 45: Please use https://github.com/hfp/libxsmm/blob/master/src/libxsmm_cpuid_x86.c#L322 and delete this routine

src/generator_mateltwise_aarch64.c
```diff
@@ -20,67 +20,70 @@
...
+  /* the A64FX has 512 bits = 64 bytes */
+  return 64;
+}
+
+LIBXSMM_API_INTERN
+void libxsmm_generator_mateltwise_aarch64_update_micro_kernel_config_vectorlength( libxsmm_generated_code*   io_generated_code,
                                                                            libxsmm_mateltwise_kernel_config* io_micro_kernel_config,
                                                                            const libxsmm_meltw_descriptor*   i_mateltwise_desc) {
-  if ( io_generated_code->arch  == LIBXSMM_AARCH64_V81 || io_generated_code->arch  == LIBXSMM_AARCH64_V82 || io_generated_code->arch  == LIBXSMM_AARCH64_APPL_M1 ) {
```
alheinecke, line 26: Please keep this, it might look redundant right now, but we might add some logic/operation with is more datatype dependent in future, hence this is aligned with the other code generators.

src/generator_mateltwise_aarch64.h
```diff
@@ -6,14 +6,22 @@
...
-/* Evanelos Georganas, Alexander Heinecke (Intel Corp.)
+/* Evanelos Georganas, Alexander Heinecke (Intel Corp.), Antonio Noack (FSU Jena)
 ******************************************************************************/
 
 #ifndef GENERATOR_MATELTWISE_AARCH64_H
 #define GENERATOR_MATELTWISE_AARCH64_H
 
 #include "generator_common.h"
 
+LIBXSMM_API_INTERN
```
alheinecke, line 17: see above, please use the LIBXSMM-wide service routines for this.

src/generator_mateltwise_unary_binary_aarch64.c
```diff
@@ -467,124 +417,51 @@ void libxsmm_load_aarch64_2d_reg_block( libxsmm_generated_code*
...
-#if 0
-  char vname = (i_vlen * i_micro_kernel_config->datatype_size_out == 64) ? 'z' : 'y';
-#endif
   unsigned int bcast_row = (((i_mateltwise_desc->operation == LIBXSMM_MELTW_OPERATION_UNARY) && ((i_mateltwise_desc->flags & LIBXSMM_MELTW_FLAG_UNARY_BCAST_ROW) > 0))) ? 1 : 0;
   unsigned int bcast_col = (((i_mateltwise_desc->operation == LIBXSMM_MELTW_OPERATION_UNARY) && ((i_mateltwise_desc->flags & LIBXSMM_MELTW_FLAG_UNARY_BCAST_COL) > 0))) ? 1 : 0;
   unsigned int bcast_scalar = (((i_mateltwise_desc->operation == LIBXSMM_MELTW_OPERATION_UNARY) && ((i_mateltwise_desc->flags & LIBXSMM_MELTW_FLAG_UNARY_BCAST_SCALAR) > 0))) ? 1 : 0;
   unsigned int l_ld_bytes = i_mateltwise_desc->ldo * i_micro_kernel_config->datatype_size_out;
   unsigned int l_m_adjust = ( i_mask_last_m_chunk == 0 ) ? i_micro_kernel_config->datatype_size_in * i_vlen * i_m_blocking : i_micro_kernel_config->datatype_size_in * ( (i_vlen * (i_m_blocking-1)) + i_mask_last_m_chunk );
   LIBXSMM_UNUSED(i_mask_reg);
-#if 0
```
alheinecke, line 492: Please don't delete #if 0 stuff this marks where we still have opens compared to x86

src/generator_mateltwise_unary_binary_aarch64.c
```diff
@@ -219,45 +219,34 @@ void adjust_in_microkernel_addr_aarch64_gp_reg( libxsmm_generated_code*
 
 LIBXSMM_API_INTERN
 void libxsmm_generator_configure_aarch64_vlens(const libxsmm_meltw_descriptor* i_mateltwise_desc, libxsmm_mateltwise_kernel_config* i_micro_kernel_config) {
+
   /* First, determine the vlen compute based on the  add the architecture check in here D-, move it to the microkernel config */
-  if ( LIBXSMM_DATATYPE_I8 == LIBXSMM_GETENUM_INP( i_mateltwise_desc->datatype2 )) {
```
alheinecke, line 223: I guess with these changes all the unit test in sample/eltwise pass for ASIMD

src/generator_mateltwise_unary_binary_aarch64.c
```diff
@@ -364,41 +351,13 @@ void libxsmm_load_aarch64_2d_reg_block( libxsmm_generated_code*
 
   /* In this case we don't have to load any data  */
   if ((i_mateltwise_desc->operation == LIBXSMM_MELTW_OPERATION_UNARY) && (i_mateltwise_desc->param == LIBXSMM_MELTW_TYPE_UNARY_XOR)) return;
-#if 0
```
alheinecke, line 367: Please don't delete #if 0 stuff this marks where we still have opens compared to x86

src/generator_mateltwise_unary_binary_aarch64.c
```diff
@@ -364,41 +351,13 @@ void libxsmm_load_aarch64_2d_reg_block( libxsmm_generated_code*
...
-#endif
   for (in = 0; in < i_n_blocking; in++) {
     for (im = 0; im < i_m_blocking; im++) {
       cur_vreg = i_start_vreg + in * i_m_blocking + im;
       if ( bcast_input == 0) {
         offset = (l_ld_bytes*i_n_blocking);
         libxsmm_generator_vloadstore_masked_vreg_aarch64_asimd( io_generated_code, i_gp_reg_mapping->gp_reg_in, i_gp_reg_mapping->gp_reg_scratch_1, cur_vreg,
                                                                 i_micro_kernel_config->datatype_size_in, (im == i_m_blocking - 1) ? i_mask_last_m_chunk : 0, 1, 0 );
-        /* If compute is in F32 and input is BF16 (or input is BF16 and output is F32), then upconvert BF16 -> FP32 */
-#if 0
```
alheinecke, line 396: Please don't delete #if 0 stuff this marks where we still have opens compared to x86

src/generator_mateltwise_unary_binary_aarch64.c
```diff
@@ -431,15 +390,6 @@ void libxsmm_load_aarch64_2d_reg_block( libxsmm_generated_code*
             libxsmm_generator_vloadstore_masked_vreg_aarch64_asimd( io_generated_code, i_gp_reg_mapping->gp_reg_in, i_gp_reg_mapping->gp_reg_scratch_1, cur_vreg,
                                                                     i_micro_kernel_config->datatype_size_in, (im == i_m_blocking - 1) ? i_mask_last_m_chunk : 0, 1, 0 );
           }
-
-#if 0
```
alheinecke, line 435: Please don't delete #if 0 stuff this marks where we still have opens compared to x86

src/generator_mateltwise_unary_binary_aarch64.c
```diff
@@ -467,124 +417,51 @@ void libxsmm_load_aarch64_2d_reg_block( libxsmm_generated_code*
...
+    for (im = 0; im < i_m_blocking; im++) {
+      cur_vreg = i_start_vreg + in * i_m_blocking + im;
+      /* In the XOR case we have a constnt vreg  */
+      if ((i_mateltwise_desc->operation == LIBXSMM_MELTW_OPERATION_UNARY) && (i_mateltwise_desc->param == LIBXSMM_MELTW_TYPE_UNARY_XOR)) {
+        cur_vreg = i_micro_kernel_config->zero_vreg;
+      } else {
+        if (bcast_row == 1) {
+          cur_vreg = i_start_vreg + in * i_m_blocking;
         }
-#if 0
```
alheinecke, line 557: Please don't delete #if 0 stuff this marks where we still have opens compared to x86

alheinecke (changes requested):

samples/sve/sve.cpp
AntonioNoack: My lacking knowledge of LIBXSMM 😅.
I'll take a look at it. What is the best way to run then with different architectures? (as I need to run them with ASIMD and SVE)

Besides that they were personal tests (they also are just part of samples, not src or include), and e.g. C has no good default for a timing library, while C++ does.

src/generator_aarch64_instructions.c
FreddieWitherden: Handling register allocation is the compiler's job.  It should have no issue keeping `code[code_head]` in a register.  Switch statements are problematic as they very often result in a branch mispredicition.

samples/sve/sve.cpp
alheinecke: LIBXSMM_TARGET as environment variable handles this and it picks automatically the highest avaiable ISA.

libxsmm_timer is your friend

alheinecke:
I still several items of the discussion open, can we please address them, then I do another review

AntonioNoack:
@alheinecke what do you suggest regarding the precision issues of 1/x? I found the same issues for 1/sqrt.
 Both functions currently use an estimation-only function with ASIMD, and have relative errors of about 5% without Newton iterations.

For SVE, I've implemented a single iteration for 1/x, which gives close to perfect results.
For 1/sqrt, the error roughly reduces by 50x with every iteration (fp32; <= 3 iterations, 4th doesn't bring improvements).
 
 What is the target for the relative error within LIBXSMM? Or should the precision be user-configurable maybe?

src/generator_aarch64_instructions.h
AntonioNoack: I'm not completely sure, what you mean by that. Is this more of a task for me or for you?

AntonioNoack:
I hope most things are resolved now :)

src/generator_aarch64_instructions.h
alheinecke: This is just a head-up that we will get some merge conflicts soon.
@FreddieWitherden : would it be doable to pull the encoder changes from the branch into a PR?

samples/sve/sve_performance_test.c
```diff
@@ -0,0 +1,132 @@
+/******************************************************************************
```
alheinecke, line 1: Couldn't we add the performance test to the correctness test in eltwise?

alheinecke (commented):
Can we move the performance tester to the eltwise perf tester, that would also benefit x86 and neon.

alheinecke:
Regarding Newton iterations: even x86 doesn't use all bits... I think we need a comparison between all of them (right now we just use approximations on all and we are even not numerical compatible between Intel and AMD chips) but for DL applications we are fine. We have Flags, so we could specify ULP here. That would be exactly what you suggest with user defined. That means if the flag is not set, it's just approximation, and then we can specify number Newton steps or ULP etc. (needs to be defined).

src/generator_aarch64_instructions.h
alheinecke: Thanks! Merge into aarch64_eltwise. @AntonioNoack: please merge aarch64_eltwise in your branch you are using for this PR.

AntonioNoack: I merged it :)
The most tedious part was to merge libxsmm_aarch64_instruction_sve_compute, but it should work for both SVE teams now.

samples/sve/sve_performance_test.c
AntonioNoack: Yes, I moved it there, and now it supports all the other architectures. However, I haven't added the feature to specify the kernels yourself yet.

AntonioNoack:
For the iterations/ulp precision, I've added the function `libxsmm_get_ulp_precision()` to generator_common.c. I've placed it into common, because other kernels and other architectures might use it as well.
The function uses the newly defined flag LIBXSMM_ULP_PRECISION, which would be an estimated target precision.

As a default, I've set 1e4 currently, which kind-of expresses that the default implementation doesn't care that much about high precision. I'm just not that sure about that magic value. Should I/we use infinity instead?

In my mind, infinity just would be a little awkward, because it might imply that the user doesn't care about the result at all (which would be false ofc, but idk)

src/generator_aarch64_instructions.h
```diff
@@ -357,14 +357,38 @@
...
+#define LIBXSMM_AARCH64_INSTR_SVE_FRSQRTS_V      0x65001c03 /* used for Newton step to improve reciprocal sqrt, unpredicated: (3-(src0*src1))/2 */
+/* define predicated SVE compute instructions */
+#define LIBXSMM_AARCH64_INSTR_SVE_FNEG_V_P       0x041da082 /* negate, vectors, predicated */
+#define LIBXSMM_AARCH64_INSTR_SVE_FSQRT_V_P      0x650da082 /* square root, vectors, predicated */
+#define LIBXSMM_AARCH64_INSTR_SVE_FMUL_V_P       0x65028082 /* multiply, vectors, predicated, src0 == dst */
+#define LIBXSMM_AARCH64_INSTR_SVE_FDIV_V_P       0x650d8082 /* divide, vectors, predicated, src0 == dst */
+#define LIBXSMM_AARCH64_INSTR_SVE_FADD_I_P       0x65188082 /* add immediate; src1 == 0 -> 0.5, else 1.0, vectors, predicated */
+#define LIBXSMM_AARCH64_INSTR_SVE_FMLA_V_P       0x65200083 /* fused multiply-add, vectors, predicated */
+#define LIBXSMM_AARCH64_INSTR_SVE_FMLS_V_P       0x65202083 /* fused multiply-subtract, vectors, predicated */
+#define LIBXSMM_AARCH64_INSTR_SVE_FMLA_V LIBXSMM_AARCH64_INSTR_SVE_FMLA_V_P /* todo: change name to fmla_v_p everywhere */
```
alheinecke, line 384: It would be goof to get that changed before the merge

src/generator_common.c
```diff
@@ -1207,3 +1207,12 @@ LIBXSMM_API_INTERN unsigned int libxsmm_compute_equalized_blocking(
   return l_ret;
 }
 
+LIBXSMM_API_INTERN float libxsmm_get_ulp_precision(void){
+  static float precision = 0;
+  if(precision <= 0){
+    char* env = getenv("LIBXSMM_ULP_PRECISION");
+    if(env) precision = (float) atof(env);
+    if(precision <= 0) precision = 1e4;/* the default precision is just an estimate; could be changed */
```
alheinecke, line 1215: Why 1e4? ULP is normally 0.5 or 1 etc... but 1e4 sounds strange

alheinecke (changes requested):
Some minor comments and I would prefer to integrating the benchmark directly into the unary correctness tester

src/generator_common.c
AntonioNoack: Because you said "That means if the flag is not set, it's just approximation, and then we can specify number Newton steps or ULP etc.", and I interpreted as performance > accuracy is the default target for LIBXSMM.
Therefore, the default setting should prefer performance, and therefore I set the expected error in ULPs to be large.

alheinecke: my point was I don't understand the 1e4, ULPs would be 0.5, 0.5-1 and sloppy or so.

AntonioNoack: If we would only define the 3 states "0.5", "1" and "sloppy", functions, which need many iterations, would be limited to 3 different numbers of iterations (one for each state).

I introduced it as a float to make it possible to support more fine-grained decisions, e.g. so you could indirectly define to use two newton-iterations for sqrt. sloppy would use no extra iterations, 0.5 and 1 would use 4 and 3 (or maybe both 4).

But I understand that probably nobody would be in that niche to require 2 iterations, and not want a setting for every kernel type.

AntonioNoack: I've changed it to an enum now, and named the "sloppy" state "estimate" :)

samples/eltwise/eltwise_binary_simple.c
```diff
@@ -14,6 +14,8 @@
 #include <stdio.h>
 #include <math.h>
 
+#include "eltwise_perf_tester.c"
+
```
alheinecke, line 18: can please make this a *h file

samples/eltwise/eltwise_binary_simple.c
```diff
@@ -321,12 +325,24 @@ int test_binary_op_f32_f32( libxsmm_blasint M, libxsmm_blasint N, libxsmm_blasin
...
+#undef BENCHMARK_FLOPS_PER_ITERATION
+#undef BENCHMARK_BANDWIDTH_PER_ITERATION
+#undef BENCHMARKED_CALL
+/* todo: isn't matmul a gemm? what is it's flops here? */
+#define BENCHMARK_FLOPS_PER_ITERATION (M * N * (binary_type == LIBXSMM_MELTW_TYPE_BINARY_MULADD ? 2 : 1))
+/* todo: would muladd/mulsub count as 3x or 4x bandwidth? */
+#define BENCHMARK_BANDWIDTH_PER_ITERATION (M * N * 3 * sizeof(float))
+#define BENCHMARKED_CALL binary_kernel(&binary_param);
+  BENCHMARK_RUN();
```
alheinecke, line 338: Why are these defines and not arguments?

samples/eltwise/eltwise_perf_tester.c
```diff
@@ -0,0 +1,138 @@
+/******************************************************************************
```
alheinecke, line 1: I put some comments on this (like the defines into the other code)

samples/eltwise/eltwise_unary_simple.c
```diff
@@ -6,14 +6,16 @@
 * Further information: https://github.com/hfp/libxsmm/                        *
```
alheinecke, line 6: see binary for comments

alheinecke (changes requested):
What's the overall status on this PR, e.g. for the functions added do all broadcast options work etc?

AntonioNoack:
No broadcasts don't work yet fully, as I have focused on getting this pull-request through.

samples/eltwise/eltwise_binary_simple.c
AntonioNoack: Thanks, I just didn't think of them, will be adjusted soon.

AntonioNoack: Fixed it :)

AntonioNoack:
Broadcasts work now :)

AntonioNoack:
The currently implemented functions are:
unary: set-zero, copy, inc, x², 1/x, sqrt, 1/sqrt
binary: add, sub, mul, div

src/generator_common_aarch64.c
```diff
@@ -165,35 +196,53 @@ void libxsmm_generator_bcastload_masked_vreg_aarch64_asimd( libxsmm_generated_co
...
+        libxsmm_aarch64_instruction_asimd_struct_r_move( io_generated_code, LIBXSMM_AARCH64_INSTR_ASIMD_LD1R_R_POST,
+                                                        i_gp_reg_addr, LIBXSMM_AARCH64_GP_REG_XZR, i_vec_reg,
+                                                        (i_datatype_size == 4) ?  LIBXSMM_AARCH64_ASIMD_TUPLETYPE_4S : LIBXSMM_AARCH64_ASIMD_TUPLETYPE_2D );
+      }
+      if ( i_masked_elems != 0 ) {
+        libxsmm_aarch64_instruction_alu_set_imm64( io_generated_code, i_gp_reg_scratch, 0x0 );
+        /* todo: this is currently only implemented for fp32; the asimd_widths need to be adjusted for other types */
+        if ( i_masked_elems == 2 ) {
+          libxsmm_aarch64_instruction_asimd_gpr_move( io_generated_code, LIBXSMM_AARCH64_INSTR_ASIMD_MOV_G_V,
+                                                      i_gp_reg_scratch, i_vec_reg, 2, LIBXSMM_AARCH64_ASIMD_WIDTH_D );
```
alheinecke, line 232: This index is out of bound, we have 2D per register, so it can only be 0/1... I just found it because it was labeled as bug fix, However, now it's out of bound and current branch passes (the basis of the PR)

src/generator_common_aarch64.c
```diff
@@ -165,35 +196,53 @@ void libxsmm_generator_bcastload_masked_vreg_aarch64_asimd( libxsmm_generated_co
...
+        } else if ( i_masked_elems == 3 ) {
+          libxsmm_aarch64_instruction_asimd_gpr_move( io_generated_code, LIBXSMM_AARCH64_INSTR_ASIMD_MOV_G_V,
+                                                      i_gp_reg_scratch, i_vec_reg, 3, LIBXSMM_AARCH64_ASIMD_WIDTH_S );
```
alheinecke, line 235: If the 2 would be correct here... than the 3 is also incorrect here.... so please revisit the change to 2 above

src/generator_common_aarch64.c
```diff
@@ -1176,32 +1235,34 @@ void libxsmm_generator_prepare_coeffs_exp_ps_3dts_aarch64( libxsmm_generated_cod
...
-  libxsmm_aarch64_instruction_broadcast_scalar_to_vec ( io_generated_code, i_vec_log2e, i_gp_reg_tmp, i_tupletype , vec_log2e );
-  libxsmm_aarch64_instruction_broadcast_scalar_to_vec ( io_generated_code, i_vec_c0, i_gp_reg_tmp, i_tupletype , vec_c0 );
-  libxsmm_aarch64_instruction_broadcast_scalar_to_vec ( io_generated_code, i_vec_c1, i_gp_reg_tmp, i_tupletype , vec_c1  );
-  libxsmm_aarch64_instruction_broadcast_scalar_to_vec ( io_generated_code, i_vec_c2, i_gp_reg_tmp, i_tupletype , vec_c2 );
-  libxsmm_aarch64_instruction_broadcast_scalar_to_vec ( io_generated_code, i_vec_c3, i_gp_reg_tmp, i_tupletype , vec_c3 );
-  libxsmm_aarch64_instruction_broadcast_scalar_to_vec ( io_generated_code, i_vec_expmask, i_gp_reg_tmp, i_tupletype , vec_expmask );
-  libxsmm_aarch64_instruction_broadcast_scalar_to_vec ( io_generated_code, i_vec_hi_bound, i_gp_reg_tmp, i_tupletype , vec_hi_bound );
-  libxsmm_aarch64_instruction_broadcast_scalar_to_vec ( io_generated_code, i_vec_lo_bound, i_gp_reg_tmp, i_tupletype , vec_lo_bound );
+  libxsmm_aarch64_instruction_broadcast_scalar_to_vec ( io_generated_code, i_vec_halves, i_gp_reg_tmp, i_tupletype, i_sve_type, i_pred_reg, vec_halves );
```
alheinecke, line 1238: I don't think we need the i_sve_type field, io_generated_code->arch carries this information

alheinecke: This is comment is for all i_sve_type in this PR, not on this function, e.g. tuple type should help to derive SVE type

alheinecke (changes requested):
Left some comments, most importantly I believe the bugfix for ASIMD introduces a bug

src/generator_common_aarch64.c
AntonioNoack: Ah ok, it makes sense now :). I'll revert it back.

src/generator_common_aarch64.c
AntonioNoack: "I don't think we need the i_sve_type field, io_generated_code->arch carries this information" No, it does not, except if SVE only targets FP32. I think SVE should also support FP64.

AntonioNoack: "tuple type should help to derive SVE type" yes, that would be possible, but I don't like the idea too much that I use a type, which was designed for ASIMD, for SVE.

This type has values, which don't make too much sense for SVE. E.g. there is 8B/16B, 4H/8H, 2S/4S, 1D/2D.
The number in front of the type could confuse somebody. It could imply that only 2 fp32s are processed, when the type is 2S.

I'd have to introduce a new function (or duplicate a lot of code) just to convert from an ASIMD type to a SVE type (when I could have used a SVE type from the start).

alheinecke:
I have just move the latest master into the branch. It has following changes:
we got rid of LIBXSMM_GEMM_PRECISION and everything is now LIBXSMM_DATATYPE, this is just very simple renaming, but has to be done in the lines coming in from PR, e.g.: LIBXSMM_GEMM_PRECISION_F32 is now LIBXSMM_DATATYPE_F32.
(for your reference the full diff of changes in master: cdb895a )

The one line typo change in your PR seems to cause some huge diff... in the complete rewrite of the x86 generator. I suggest reverting changes in the PR in order to have clean merge.

src/generator_common_aarch64.c
alheinecke: That means we need a refactoring here. Tupletype is intrinsically connected the processiong width as you correctly say (ARM is a weakly typed ISA so 4S just means that we interpret the SIMD register as 4x32bit, but not FP32). 
So yes, there are three options:
- having different functions for SVE and ASIMD as SVE doesn't need tuple types due to masking
- having some kind of infrastructure as you are proposing such that we don't have datatype and tuple as arguments.
- using only masking and deriving from the masked entries how to call the emulation code we do today on ASIMD by using different combinations of tuple types.

AntonioNoack:
I was able to merge in those changes with `git pull -X theirs https://github.com/hfp/libxsmm aarch64_eltwise` relatively easily :)

Also, before Christmas, I've added exp, dexp/dx, tanh and dtanh/dx

alheinecke:
OK, that means all except ReLU, Leaky ReLU and GeLU are working incl. broadcast support?

AntonioNoack:
I took a look at the list of enums for the unary operations, and it looks like quite a few are still missing additionally:
- Reduce_ (x op add, x² op add, x_x² op add, x op max, x op mul, x op add ncnc format, to scalar op add),
- Dropout,
- Replicate ColVar,
- Transform (Norm, VNNI, PadM),
- Unpack to Blocks,
- Elu,
- Statistic Round,
- Quant,
- Dequant,
- Reduce Cols Idx

Then also there is a flag called REDUCE_XOR_ACC, and am unsure, whether it is implemented with the current functions.

AntonioNoack:
And yes, all others should be working including broadcasting.

AntonioNoack:
Hi, I wanted to implemented GeLU next,
but I don't understand a part of the existing ASIMD assembly:
```
    libxsmm_aarch64_instruction_asimd_compute( io_generated_code, LIBXSMM_AARCH64_INSTR_ASIMD_SHL_I_V,
                                               i_vec_index, LIBXSMM_AARCH64_ASIMD_REG_UNDEF, 2, i_vec_index,
                                               i_tupletype );

    libxsmm_aarch64_instruction_asimd_compute( io_generated_code, LIBXSMM_AARCH64_INSTR_ASIMD_TBL_1,
                                               i_vec_index, i_vec_tmp, 0, i_vec_index,
                                               i_tupletype );
```

To me this looks like
```
i_vec_index = i_vec_index << 2; // or i_vec_index *= 4;
i_vec_index = i_vec_tmp[i_vec_index];// as far as I understand TBL_1, it picks the i_vec_index-th element from the vector. If the index is out of bounds, so > 3, it would set the value to 0 instead.
```
As I understand it, this itself only would pick element 0, or would it set the result to zero, because in ASIMD fp32, there are only 4 elements in a vector.

My issue is that i_vec_tmp seems to be initialized to { 0x00000000, 0x04040404, 0x08080808, 0x0c0c0c0c }, so the result always would be zero.

Where is my mistake in understanding this? This would be too much effort to just set a value to zero.

AntonioNoack:
Or is it maybe using the index as a byte index? 🤔, then it would work, and effectively be a index < 3 ? index *4 : 0.

alheinecke:
Yes the tbl instruction is working on bytes and if I recall this code correctly it's converting 32 bit indecies into 32 bit values tables to the same table but organized as bytes values with bytes indices
________________________________

AntonioNoack:
I have a question about the type RELU_INV. Usually, the inv-type is the derivative of the function, but for RELU, it's not (based on the implementation and the test).

Is this intended behavior or a bug?
The implementation says x < 0 ? 0 : x, while the true derivative is x < 0 ? 0 : 1.
(I can imagine that using this instead of the true derivative gives a better gradient descent behavior, but it still might be unexpected for some backpropagation implementations)

alheinecke:
The derivative you mention for ReLU is correct, however RELU_INV is not the derivative of ReLU it is the backward prop. Here we multiply the gradients with the derivative of ReLU from forward prop. We just use x for activations in FWD and gradient activation in BWD (sloppy notation/variable names). Therefore for BWD ReLU, we multiply 1*gradient if gradient > 0 and 0 otherwise. 

oneDNN has a nice summary from activation and gradient perspective:
https://oneapi-src.github.io/oneDNN/dev_guide_eltwise.html#general

AntonioNoack:
I fixed the merge conflicts, that emerged by other updates to the aarch64 branch

alheinecke (commented):
As mentioned in the issue you opened and after reviewing this one. It might be good to open 2 PRs against master ASAP:
a) The CPUID changes and the broken incl. (I can also fix with a commit if you don't want credit for it)
b) the benchmarking infrastructure for eltwise. We have more changes coming as we make all test weakly typed right now. So there is another round of changes needed. We hope we have the weakly typed eltwise tester in master by Sunday.

alheinecke:
are we ready to resolve the conflicts and attempt a merge?

AntonioNoack:
Yes :). I already did, but I am not yet 100% sure that everything still works. It compiles, and sqrt still works (so maybe it's fine).

AntonioNoack:
About my performance tests, I am not really sure. Currently, because there was tons of changes in the correctness checks, I just replaced them with what was in the aarch64_eltwise branch. (except that I didn't delete eltwise_perf_tester.h yet)

AntonioNoack:
For example I saw https://github.com/libxsmm/libxsmm/pull/597/commits/abe03b55851f375f53718676b32f4799cc1b1b15, and it would have to be integrated back in again.
Which probably means that the switch-case statement should become a else-if cascade again too.

AntonioNoack:
I've fixed the tanh issue, and the switch-case->else-if issue.
I also added the tests back in, but as a part of eltwise_common.h like discussed.

I know the tests would have been better as a separate pull request, but it's quite tedious to find out which exact commands to use to create a branch from a remote repo 😅 (and unfortunately, last time I didn't write it down). If it needs to be separate, so you can merge the SVE stuff into aarch64_eltwise and the tests into main, I could take a look at it as well.

samples/eltwise/eltwise_opreduce_idxvecs.c
```diff
@@ -390,73 +382,62 @@ int main(int argc, char* argv[])
         vec_in = inp_matrix2;
       }
 
-      if (op != OP_COPY) {
```
alheinecke, line 393: I"m not sure if I understand the point of all these changes, can you elaborate, please?

src/generator_aarch64_instructions.c
```diff
@@ -1019,54 +1024,160 @@ void libxsmm_aarch64_instruction_sve_compute( libxsmm_generated_code*        io_
...
       exit(-1);
+    }
+  }
+
+  /* special instruction, where only dst = src_0 is supported */
+  /* this check could be disabled for performance reasons */
+  if( (l_vec_instr & LIBXSMM_AARCH64_INSTR_SVE_IS_DESTRUCTIVE) == LIBXSMM_AARCH64_INSTR_SVE_IS_DESTRUCTIVE ){
+    if( i_vec_reg_src_0 != i_vec_reg_dst ){
+      if(i_vec_reg_src_1 == i_vec_reg_dst &&
+        (l_vec_instr == LIBXSMM_AARCH64_INSTR_SVE_FMAX_V_P ||
```
alheinecke, line 1111: Is there a way to avoid checking the instruction type?

src/generator_aarch64_instructions.c
```diff
@@ -1019,54 +1024,160 @@ void libxsmm_aarch64_instruction_sve_compute( libxsmm_generated_code*        io_
...
+         l_vec_instr == LIBXSMM_AARCH64_INSTR_SVE_FMIN_V_P ||
+         l_vec_instr == LIBXSMM_AARCH64_INSTR_SVE_FMUL_V_P)){
+        /* arguments can be switched around */
+        /* we assign 0 <- 1 anyways, so we just skip that */
+      } else if(i_vec_reg_src_1 == i_vec_reg_dst &&
+        (l_vec_instr == LIBXSMM_AARCH64_INSTR_SVE_FDIV_V_P ||
```
alheinecke, line 1117: dito

alheinecke (commented):
Please address the minor issues and please put a summary comment what yet doesn't on SVE

src/generator_aarch64_instructions.c
AntonioNoack: Kind of.
We can force the programmer to use the correct argument order, and just crash if it is incorrect.

But I'd prefer if we detect that it doesn't matter (for div that another instruction can be used), and continue.

In the reduce kernels, there is a lot of calls to these instructions, where ASIMD didn't care about argument order. It might be fixable in those kernels, but every case would have to be checked.

src/generator_aarch64_instructions.c
AntonioNoack: dito (same arguments as fmul,fmax,fmin)

samples/eltwise/eltwise_opreduce_idxvecs.c
```diff
@@ -381,7 +373,7 @@ int main(int argc, char* argv[])
     }
     for (_i = 0; _i < m; _i++) {
       i = _i;
-      if (bcast_factor > 0) {
+      if (bcast_factor > 1) {
```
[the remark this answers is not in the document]
AntonioNoack: No, it is a beautification or simplification.
We don't need to divide by 1.

AntonioNoack:
Summary on what is (or should be) working:

**Unary:**
- copy
- xor (set zero)
- inc
- neg
- x²
- 1/x
- 1/sqrt(x)
- sqrt(x)
- tanh, tanh-inv
- sigmoid, sigmoid-inv
- exponential
- gelu, gelu-inv (currently only on 512 bit vector length; should be relatively easy to adapt with TBX)
- elu, elu-inv
- relu, relu-inv
- leaky-relu, leaky-relu-inv
- dropout, dropout-inv

**Binary:**
- add
- mul
- sub
- div
- muladd

Reduce:
- sum, sum-squared, max

For all kernels, SVE should have the same support for additional flags, and should support everything that ASIMD supports.

Examples, on what neither ASIMD nor SVE support:
- reduction with MIN (should be relatively easy to add, if needed)
- binary type DOT
- double precision in some cases

samples/eltwise/eltwise_opreduce_idxvecs.c
AntonioNoack: The purpose of those is just superficial. The old if-else-layout was something like
```
if(!copy){
  if(add) { ... }
  if(mul){ ... }
  if(max){ ... }
} else { copy }
```
and I replaced it with
```
if(copy){}
else if(add){}
else if(mul){}
else if(max){}
```
These branches are all mutually exclusive, and I think my layout better represents that "copy" is just a type like all others (e.g. "mul" or "add"). The first one looks like "copy" is special, and that multiple cases might be true (which is false).

alheinecke:
Thanks for your contribution!

PR #564, opened 2021-11-28
Enabling transpose of A in GEMM via an intermediate transpose on the stack

kvoronin-intel:
Summary:
Adding a stack transpose as a means to enable GEMM with trans_a = 1. For transposing on the stack, meltwise transform generator is used. Currently only for SP, AVX512.

Note:
A not-really-clean version to initiate a code review and understand what should be the desired final state (hoisting the changes to libxsmm_generator_gemm_kernel if we want to enable trans_a for a wider configuration scope?).

Testing:
Code has been only tested for functional correctness on just a couple of problems with square and rectangular A. More will be needed before merging.

### DIFF
--- a/include/libxsmm_generator.h
+++ b/include/libxsmm_generator.h
@@ -14,10 +14,8 @@
 #include "libxsmm_typedefs.h"
 
 #define LIBXSMM_GEMM_NO_BYPASS(FLAGS, ALPHA, BETA) ( \
-  0 == ((FLAGS) & (LIBXSMM_GEMM_FLAG_TRANS_A)) && \
-        (LIBXSMM_FEQ(1, ALPHA) /*|| LIBXSMM_FEQ(-1, ALPHA)*/) && \
-        (LIBXSMM_FEQ(1, BETA) || LIBXSMM_FEQ(0, BETA)))
-
+  0 ==  (LIBXSMM_NEQ(1, ALPHA) /*&& LIBXSMM_NEQ(-1, ALPHA)*/) || \
+        (LIBXSMM_NEQ(1, BETA) && LIBXSMM_NEQ(0, BETA)))
 
 /** Initialize GEMM descriptor as used by low-level routines (type-specific). */
 LIBXSMM_API libxsmm_gemm_descriptor* libxsmm_dgemm_descriptor_init(libxsmm_descriptor_blob* blob,
--- a/samples/xgemm/kernel.c
+++ b/samples/xgemm/kernel.c
@@ -136,11 +136,23 @@ void ref_matmul( gemm_def* i_gemm_def, void* a, void* b, void* c ) {
             f_c[(l_j * ldc) + l_i] = 0.0;
           }
           for (l_s = 0; l_s < k; l_s++) {
-            if ( i_gemm_def->trans_b == 0 ) {
-              f_c[(l_j * ldc) + l_i] += f_a[(l_r * lda * k) + ((l_s * lda) + l_i)] * f_b[(l_r * ldb * n) + ((l_j * ldb) + l_s)];
+            if (i_gemm_def->trans_b == 0) {
+              if (i_gemm_def->trans_a == 0) {
+                f_c[(l_j * ldc) + l_i] += f_a[(l_r * lda * k) + (l_s * lda) + l_i] *
+                                                   f_b[(l_r * ldb * n) + (l_j * ldb) + l_s];
+              } else {
+                f_c[(l_j * ldc) + l_i] += f_a[(l_r * lda * m) + (l_i * lda) + l_s] *
+                                                   f_b[(l_r * ldb * n) + (l_j * ldb) + l_s];
+              } // if-else l_trans_a
             } else {
-              f_c[(l_j * ldc) + l_i] += f_a[(l_r * lda * k) + ((l_s * lda) + l_i)] * f_b[(l_r * ldb * k) + ((l_s * ldb) + l_j)];
-            }
+              if (i_gemm_def->trans_a == 0) {
+                f_c[(l_j * ldc) + l_i] += f_a[(l_r * lda * k) + (l_s * lda) + l_i] *
+                                                   f_b[(l_r * ldb * k) + (l_s * ldb) + l_j];
+              } else { // untested?
+                f_c[(l_j * ldc) + l_i] += f_a[(l_r * lda * m) + (l_i * lda) + l_s] *
+                                                   f_b[(l_r * ldb * k) + (l_s * ldb) + l_j];
+              } // if-else l_trans_a
+            } // if-else l_trans_b
           }
         }
       }
@@ -431,8 +443,7 @@ double jit_matmul( const gemm_def*    i_gemm_def,
     l_flags |= LIBXSMM_GEMM_FLAG_TRANS_B;
   }
   if ( i_gemm_def->trans_a != 0 ) {
-    fprintf(stderr, "trans_a needs to be 0\n");
-    return EXIT_FAILURE;
+    l_flags |= LIBXSMM_GEMM_FLAG_TRANS_A;
   }
   if ( i_gemm_def->vnni_a != 0 ) {
     l_flags |= LIBXSMM_GEMM_FLAG_VNNI_A;
@@ -1010,7 +1021,11 @@ int main(int argc, char* argv []) {
     l_c_perf = (char*)libxsmm_aligned_malloc((size_t)l_ldc * (size_t)l_n * LIBXSMM_TYPESIZE(l_gemm_def.out_type), 64);
     l_c_gold = (char*)libxsmm_aligned_malloc((size_t)l_ldc * (size_t)l_n * LIBXSMM_TYPESIZE(l_gemm_def.out_type), 64);
 
-    init_random_matrix( l_gemm_def.in_type, l_a, l_br, l_lda, l_k );
+    if (l_gemm_def.trans_a == 0) {
+      init_random_matrix( l_gemm_def.in_type, l_a, l_br, l_lda, l_k );
+    } else {
+      init_random_matrix( l_gemm_def.in_type, l_a, l_br, l_lda, l_m );
+    }
     if (l_gemm_def.trans_b == 0) {
       init_random_matrix( l_gemm_def.in_type, l_b, l_br, l_ldb, l_n );
     } else {
--- a/samples/xgemm/kernel.c
+++ b/samples/xgemm/kernel.c
@@ -1011,7 +1011,11 @@ int main(int argc, char* argv []) {
     l_gemm_def.ldb = l_ldb;
     l_gemm_def.ldc = l_ldc;
 
-    l_a      = (char*)libxsmm_aligned_malloc((size_t)l_lda * (size_t)l_k * (size_t)l_br * LIBXSMM_TYPESIZE(l_gemm_def.in_type), 64);
+    if (l_gemm_def.trans_a == 0) {
+      l_a      = (char*)libxsmm_aligned_malloc((size_t)l_lda * (size_t)l_k * (size_t)l_br * LIBXSMM_TYPESIZE(l_gemm_def.in_type), 64);
+    } else {
+      l_a      = (char*)libxsmm_aligned_malloc((size_t)l_lda * (size_t)l_m * (size_t)l_br * LIBXSMM_TYPESIZE(l_gemm_def.in_type), 64);
+    }
     if (l_gemm_def.trans_b == 0) {
       l_b      = (char*)libxsmm_aligned_malloc((size_t)l_ldb * (size_t)l_n * (size_t)l_br * LIBXSMM_TYPESIZE(l_gemm_def.in_type), 64);
     } else {
--- a/samples/xgemm/kernel_test/sgemm_tn_eqld.slurm
+++ b/samples/xgemm/kernel_test/sgemm_tn_eqld.slurm
@@ -31,21 +31,21 @@ f1.close()
 f2.close()
 END
 
-./kernel ${TESTFILE1} 1 1 0 0 1 0 SP nobr 1 0 1 1
+./kernel ${TESTFILE1} 1 1 0 0 1 0 SP nobr 1 0 1 0
 
-./kernel ${TESTFILE2} 1 1 0 0 1 0 SP nobr 1 0 1 1
+./kernel ${TESTFILE2} 1 1 0 0 1 0 SP nobr 1 0 1 0
 
-./kernel ${TESTFILE1} 1 1 0 0 1 0 SP addrbr 5 0 1 1
+./kernel ${TESTFILE1} 1 1 0 0 1 0 SP addrbr 5 0 1 0
 
-./kernel ${TESTFILE1} 1 1 0 0 1 0 SP offsbr 5 0 1 1
+./kernel ${TESTFILE1} 1 1 0 0 1 0 SP offsbr 5 0 1 0
 
-./kernel ${TESTFILE1} 1 1 0 0 1 0 SP strdbr 5 0 1 1
+./kernel ${TESTFILE1} 1 1 0 0 1 0 SP strdbr 5 0 1 0
 
-./kernel ${TESTFILE2} 1 1 0 0 1 0 SP addrbr 5 1 1 1
+./kernel ${TESTFILE2} 1 1 0 0 1 0 SP addrbr 5 1 1 0
 
-./kernel ${TESTFILE2} 1 1 0 0 1 0 SP offsbr 5 1 1 1
+./kernel ${TESTFILE2} 1 1 0 0 1 0 SP offsbr 5 1 1 0
 
-./kernel ${TESTFILE2} 1 1 0 0 1 0 SP strdbr 5 1 1 1
+./kernel ${TESTFILE2} 1 1 0 0 1 0 SP strdbr 5 1 1 0
 
 rm ${TESTFILE1}
 rm ${TESTFILE2}
--- a/samples/xgemm/kernel_test/sgemm_tn_eqld.slurm
+++ b/samples/xgemm/kernel_test/sgemm_tn_eqld.slurm
@@ -1,0 +1,4 @@
+#!/usr/bin/env bash
+
+./kernel 32 32 32 32 32 32 1 1 0 0 1 0 nopf SP nobr 1 0 1 1
+
--- a/src/generator_gemm.c
+++ b/src/generator_gemm.c
@@ -173,7 +173,10 @@ void libxsmm_generator_gemm_kernel( libxsmm_generated_code*        io_generated_
   if ( (l_xgemm_desc_mod.flags & LIBXSMM_GEMM_FLAG_TRANS_A) == LIBXSMM_GEMM_FLAG_TRANS_A ) {
     // Non FP32 and BRGEMM are not supported for the trans_a = 1 case
     if ((LIBXSMM_GEMM_PRECISION_F32 != LIBXSMM_GETENUM_INP( l_xgemm_desc_mod.datatype )) ||
-        (l_xgemm_desc_mod.flags & LIBXSMM_GEMM_FLAG_BATCH_REDUCE_ADDRESS              )) {
+        (l_xgemm_desc_mod.flags & LIBXSMM_GEMM_FLAG_BATCH_REDUCE_ADDRESS               ) ||
+        (l_xgemm_desc_mod.flags & LIBXSMM_GEMM_FLAG_BATCH_REDUCE_OFFSET                ) ||
+        (l_xgemm_desc_mod.flags & LIBXSMM_GEMM_FLAG_BATCH_REDUCE_STRIDE                )
+       ) {
       LIBXSMM_HANDLE_ERROR( io_generated_code, LIBXSMM_ERR_INVALID_GEMM_CONFIG );
       return;
     }
--- a/src/generator_gemm.c
+++ b/src/generator_gemm.c
@@ -170,7 +170,13 @@ void libxsmm_generator_gemm_kernel( libxsmm_generated_code*        io_generated_
   }
 
   /* check LDA */
-  if ( (l_xgemm_desc_mod.flags & LIBXSMM_GEMM_FLAG_TRANS_A) > 0 ) {
+  if ( (l_xgemm_desc_mod.flags & LIBXSMM_GEMM_FLAG_TRANS_A) == LIBXSMM_GEMM_FLAG_TRANS_A ) {
+    // Non FP32 and BRGEMM are not supported for the trans_a = 1 case
+    if ((LIBXSMM_GEMM_PRECISION_F32 != LIBXSMM_GETENUM_INP( l_xgemm_desc_mod.datatype )) ||
+        (l_xgemm_desc_mod.flags & LIBXSMM_GEMM_FLAG_BATCH_REDUCE_ADDRESS              )) {
+      LIBXSMM_HANDLE_ERROR( io_generated_code, LIBXSMM_ERR_INVALID_GEMM_CONFIG );
+      return;
+    }
     if ( l_xgemm_desc_mod.lda < l_xgemm_desc_mod.k ) {
       LIBXSMM_HANDLE_ERROR( io_generated_code, LIBXSMM_ERR_LDA_TRANS );
       return;
--- a/src/generator_gemm.c
+++ b/src/generator_gemm.c
@@ -171,7 +171,7 @@ void libxsmm_generator_gemm_kernel( libxsmm_generated_code*        io_generated_
 
   /* check LDA */
   if ( (l_xgemm_desc_mod.flags & LIBXSMM_GEMM_FLAG_TRANS_A) == LIBXSMM_GEMM_FLAG_TRANS_A ) {
-    // Neither non-FP32 nor BRGEMM are supported for the trans_a = 1 case
+    /* Neither non-FP32 nor BRGEMM are supported for the trans_a = 1 case */
     if ((l_xgemm_desc_mod.flags & LIBXSMM_GEMM_FLAG_BATCH_REDUCE_ADDRESS               ) ||
         (l_xgemm_desc_mod.flags & LIBXSMM_GEMM_FLAG_BATCH_REDUCE_OFFSET                ) ||
         (l_xgemm_desc_mod.flags & LIBXSMM_GEMM_FLAG_BATCH_REDUCE_STRIDE                )
--- a/src/generator_gemm.c
+++ b/src/generator_gemm.c
@@ -170,9 +170,16 @@ void libxsmm_generator_gemm_kernel( libxsmm_generated_code*        io_generated_
   }
 
   /* check LDA */
-  if ( l_xgemm_desc_mod.lda < l_xgemm_desc_mod.m ) {
-    LIBXSMM_HANDLE_ERROR( io_generated_code, LIBXSMM_ERR_LDA );
-    return;
+  if ( (l_xgemm_desc_mod.flags & LIBXSMM_GEMM_FLAG_TRANS_A) > 0 ) {
+    if ( l_xgemm_desc_mod.lda < l_xgemm_desc_mod.k ) {
+      LIBXSMM_HANDLE_ERROR( io_generated_code, LIBXSMM_ERR_LDA_TRANS );
+      return;
+    }
+  } else {
+    if ( l_xgemm_desc_mod.lda < l_xgemm_desc_mod.m ) {
+      LIBXSMM_HANDLE_ERROR( io_generated_code, LIBXSMM_ERR_LDA );
+      return;
+    }
   }
 
   /* check LDB */
@@ -192,6 +199,16 @@ void libxsmm_generator_gemm_kernel( libxsmm_generated_code*        io_generated_
   if ( l_xgemm_desc_mod.ldc < l_xgemm_desc_mod.m ) {
     LIBXSMM_HANDLE_ERROR( io_generated_code, LIBXSMM_ERR_LDC );
     return;
+  }
+
+  /* check for trans A cases which are not supported in the generator */
+  if ( (l_xgemm_desc_mod.flags & LIBXSMM_GEMM_FLAG_TRANS_A) > 0 ) {
+    if ( (LIBXSMM_DATATYPE_F32 != LIBXSMM_GETENUM_INP( l_xgemm_desc_mod.datatype )) ) {
+      LIBXSMM_HANDLE_ERROR( io_generated_code, LIBXSMM_ERR_TRANS_A );
+      return;
+    } else {
+      /* we are fine, we have transpose support */
+    }
   }
 
   /* check for trans B cases which are not supported in the generator */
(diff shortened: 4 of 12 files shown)


alheinecke:
after the github outage I have rebased the PR to the a_transpose branch on upstream (just copy of latest master for now) and I do detailed review later.

include/libxsmm_generator.h
```diff
@@ -13,11 +13,15 @@
...
-
+// Temporarily and loosely enabling transpose of A for all GEMMs
+// (while currently we don't know if it will actually be supported there)
+// Likely we need to restrict this at least to AVX512 and FP32/FP64
+// or extend the code changes to hoist the transpose up the call stack
+// In the past the check had:
+//  0 == ((FLAGS) & (LIBXSMM_GEMM_FLAG_TRANS_A)) && \
+//        (LIBXSMM_FEQ(1, ALPHA) /*|| LIBXSMM_FEQ(-1, ALPHA)*/) && \
+//        (LIBXSMM_FEQ(1, BETA) || LIBXSMM_FEQ(0, BETA)))
+#define LIBXSMM_GEMM_NO_BYPASS(FLAGS, ALPHA, BETA) (1)
```
alheinecke, line 24: I think we should only remove the transpose A check

samples/xgemm/kernel.c
```diff
@@ -491,108 +491,218 @@ int main(int argc, char* argv []) {
     } while ( l_keep_going );
   }
   else if ((strcmp(l_precision, "SP") == 0) && (l_trans_b == 0)) {
-    unsigned int l_keep_going = 0;
```
alheinecke, line 494: perhaps we should think about fixing this better than the trans_b stuff, otherwise we implement the same stuff 4x times

samples/xgemm/kernel_jit_exec.h
```diff
@@ -14,6 +14,7 @@
 #include <math.h>
 #include <string.h>
 
+#define A_TRANS_ENABLED
```
alheinecke, line 17: I guess we can remove this.

src/generator_gemm.c
```diff
@@ -170,9 +170,16 @@ void libxsmm_generator_gemm_kernel( libxsmm_generated_code*        io_generated_
   }
 
   /* check LDA */
-  if ( l_xgemm_desc_mod.lda < l_xgemm_desc_mod.m ) {
-    LIBXSMM_HANDLE_ERROR( io_generated_code, LIBXSMM_ERR_LDA );
-    return;
+  if ( (l_xgemm_desc_mod.flags & LIBXSMM_GEMM_FLAG_TRANS_A) > 0 ) {
```
alheinecke, line 173: Maybe checking against exactly: 
(l_xgemm_desc_mod.flags & LIBXSMM_GEMM_FLAG_TRANS_A) == LIBXSMM_GEMM_FLAG_TRANS_A)

Also we should check for now not only on FP32, but also that no Batch Reduce GEMM is used (also flag fields). In the moment a Batch reduce flag is set we should error out for now. It's simple to add, but we need to transpose several a matrices and mock with more inputs and pointers....

src/generator_gemm_sse_avx_avx2_avx512.c
```diff
@@ -18,31 +18,19 @@
 #include "generator_gemm_avx512_microkernel.h"
 #include "libxsmm_main.h"
 
-LIBXSMM_API_INTERN void libxsmm_generator_gemm_sse_avx_avx2_avx512_kernel( libxsmm_generated_code*        io_generated_code,
```
alheinecke, line 21: The first 100 lines look pretty messy in the diff. I think it's fine, but perhaps you double and do a commit which has a cleaner diff?

src/generator_gemm_sse_avx_avx2_avx512.c
```diff
@@ -105,21 +93,52 @@ LIBXSMM_API_INTERN void libxsmm_generator_gemm_sse_avx_avx2_avx512_kernel( libxs
...
+  if (i_xgemm_desc->flags & LIBXSMM_GEMM_FLAG_BATCH_REDUCE_ADDRESS) {
+    if ( i_xgemm_desc->prefetch == LIBXSMM_GEMM_PREFETCH_AL2          ||
+         i_xgemm_desc->prefetch == LIBXSMM_GEMM_PREFETCH_AL2BL2_VIA_C    ) {
+      adjust_A_pf_ptrs = 1;
+    }
+  }
+
+  /* in case when A needs to be transposed, we need to change temporarily the desciptor dimensions for gemm */
+  libxsmm_descriptor_blob l_blob_opa;
+  unsigned int lda_transpose = i_xgemm_desc->m;
```
alheinecke, line 119: starting here the diff gets clean.

src/generator_gemm_sse_avx_avx2_avx512.c
```diff
@@ -137,29 +156,143 @@ LIBXSMM_API_INTERN void libxsmm_generator_gemm_sse_avx_avx2_avx512_kernel( libxs
...
+#if defined(_WIN32) || defined(__CYGWIN__)
+    l_mateltwise_gp_reg_mapping.gp_reg_param_struct = LIBXSMM_X86_GP_REG_RCX;
+#else /* match calling convention on Linux */
+    l_mateltwise_gp_reg_mapping.gp_reg_param_struct = LIBXSMM_X86_GP_REG_RDI;
+#endif
+
+    /* stack management at the start (does nothing for my case?) for meltw kernel */
+    libxsmm_generator_meltw_setup_stack_frame( io_generated_code, l_mateltwise_desc, &l_mateltwise_gp_reg_mapping, &l_mateltwise_kernel_config);
+
+    libxsmm_generator_transform_norm_to_normt_32bit_avx512_microkernel( io_generated_code, &l_loop_label_tracker,
```
alheinecke, line 254: here we should add an arch-checker

alheinecke (changes requested):
LGTM, some minor comments such that x86 and FP32 and FP64 could use the feature should be added.

include/libxsmm_generator.h
kvoronin-intel: I got a bit confused by the existing check.
Independent of transpose A check, if, say, alpha != 1, then LIBXSMM_FEQ(1,alpha) will be 0 and the check will be true as we compare righthand side to 0. 
But it is used in expressions like this one
```
 if  (LIBXSMM_GEMM_NO_BYPASS(flags, alpha, beta)
    && ... )
  {
    ...
    LIBXSMM_GEMM_DESCRIPTOR(...);
  }
```
So if the check intended to bail out if alpha != 1 (which I think it was), it wasn't doing that. Of course, it can be fixed by changing FEQ to NEQ and swapping && with ||.
But am I correct here?

samples/xgemm/kernel.c
kvoronin-intel: Sure, will reduce the code duplication. I had same exact thought but looked at trans_b stuff and decided to use the same boilerplate approach.

samples/xgemm/kernel_jit_exec.h
kvoronin-intel: removed

src/generator_gemm.c
kvoronin-intel: I'll add a check for F32.
Checking for exact match with `LIBXSMM_GEMM_FLAG_TRANS_A` will exclude possibility of modifiers like LIBXSMM_GEMM_FLAG_ALIGN_A and LIBXSMM_GEMM_FLAG_ALIGN_C, LIBXSMM_GEMM_FLAG_ALIGN_C_NTS_HINT + some cobinations. 
Is it ok to disable those (by making an exact match for TRANS_A)?

samples/xgemm/kernel_jit_exec.h
kvoronin-intel: Removed

src/generator_gemm_sse_avx_avx2_avx512.c
kvoronin-intel: Removed unnecessary code shuffling so that the diff looks much cleaner.

samples/xgemm/kernel.c
kvoronin-intel: Attempted to unite the code for trans_a, as well as trans_b, in case of SP. Is this better?

src/generator_gemm_sse_avx_avx2_avx512.c
kvoronin-intel: do you want to have a dispatch here and call appropriate microkernels? i see there exist at least _sse, _avx and _spr/_pre_spr microkernels which could be called along with _avx512 at this place [unless there are subtle reasons why some of them cannot be called in the same way as _avx512 is now].

alheinecke: yes, a dispatch for sse, avx and avx512 (as it is right now). the deeper specializing for avx512 pre/post Golden Cove core is done internally.

src/generator_gemm_sse_avx_avx2_avx512.c
alheinecke: Thanks this looks very good now.

alheinecke (commented):
Thanks, very nice!

I think we only thing missing the guarding for batch reduce and the sse/avx/avx512 dispatch

include/libxsmm_generator.h
alheinecke: The check is correct now. It should bail out if alpha != 1 and if beta != 0 or != 1.... in other words we can only JIT kernel if alpha == 1 and if beta is either 0 or 1

src/generator_gemm.c
alheinecke: These are independent flags, we really only want to check that one bit.

kvoronin-intel: Right, mixed up smth aabout the flags. Added a check for FP32 and BRGEMM. Is `LIBXSMM_ERR_INVALID_GEMM_CONFIG` an appropriate error for the unsupported cases here?

src/generator_gemm_sse_avx_avx2_avx512.c
kvoronin-intel: Added a dispatch, quickly checked additionally for avx, avx2 and sse42 that code is visually correct. For p4p (after adding export LIBXSMM_TARGET=x86_64) a simple test worked as well.

kvoronin-intel:
> Thanks, very nice!
> 
> I think we only thing missing the guarding for batch reduce and the sse/avx/avx512 dispatch

Great! What are the testing guidelines in general? Is there a dedicated test suite, or standard set of testing configurations to extend for trans_a & test? So far I only added a small test in samples/xgemm

src/generator_gemm.c
alheinecke: yes!

alheinecke:
in kernel_test (in the same directory) we have some scripts which draw random dimension. They should be adopted for A Trans and then we are good.

src/generator_gemm.c
kvoronin-intel: Did a change with `LIBXSMM_ERR_INVALID_GEMM_CONFIG`.
But, when I run, e.g. the kernel_test with some flavour of BRGEMM, the error with LIBXSMM_VERBOSE=-1 would look like:
"JIT failed, please run with LIBXSMM_VERBOSE=-1 and/or with debug mode LIBXSMM library!
LIBXSMM ERROR (libxsmm_generator_gemm_kernel): invalid GEMM config in setup detected (error #90035)!"
Is it better than throwing `LIBXSMM_ERR_LDA_TRANS` which will give a more transpose-specific message
"GEMM kernel with trans A requested, but target/datatype not supported (error ...)"?

samples/xgemm/kernel_test/sgemm_tn_eqld.slurm
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/env bash
+
+TESTFILE1=$(mktemp -p .)
```
kvoronin-intel, line 3: Added a standard test for sgemm_tn, with "eq" and "gt" cases for leading dimensions.

src/generator_gemm_sse_avx_avx2_avx512.c
alheinecke: great

samples/xgemm/kernel_test/sgemm_tn_eqld.slurm
alheinecke: great

src/generator_gemm.c
alheinecke: Yeah, you are right let's add a specific error code, perhaps even just for BRGEMM. BRGEMM could use the discussed gather & and inner product version though. Later on we can just put all A_i matrices on the stack, transpose them ( \forall A'_i = trans(A_i) and then feed the new A'_i into the BRGEMM.

alheinecke (commented):
LGTM, if think after your suggestion with having a different error code we can merge this back into the a_transpose branch

src/generator_gemm.c
kvoronin-intel: Added a new error code, so that now the output for BRGEMM flavors looks like this:
`JIT failed, please run with LIBXSMM_VERBOSE=-1 and/or with debug mode LIBXSMM library!
LIBXSMM ERROR (libxsmm_generator_gemm_kernel): BRGEMM is not supported for the specified configuration with A transpose (error #90048)!`
Let me know if you had in mind a more general error like "Unsupported GEMM configuration for BRGEMM case" or smth.

alheinecke:
I have just move the latest master into the branch. It has following changes:
- new GEMM frontend, which should not matter for you: https://github.com/hfp/libxsmm/blob/master/src/generator_gemm_sse_avx_avx2_avx512.c#L174, the transpose code goes just after this new block as this is just setting up the input registers in a different way. (I also move to the open stream up into another function but that shouldn't matter).
- we got rid of LIBXSMM_GEMM_PRECISION and everything is now LIBXSMM_DATATYPE, this is just very simple renaming, but has to be done in the lines coming in from PR, e.g.: LIBXSMM_GEMM_PRECISION_F32 is now LIBXSMM_DATATYPE_F32.
(for your reference the full diff of changes in master: https://github.com/hfp/libxsmm/commit/cdb895a3cbd38b414f5420617be46132fa082155 )

alheinecke:
I have heavily refactored the kernel.c and kernel_parallel.c performance tester to be weakly typed. That removed 4K LoC Spaghetti code, so testing for trans_a should be much easier now.

kvoronin-intel:
> I have heavily refactored the kernel.c and kernel_parallel.c performance tester to be weakly typed. That removed 4K LoC Spaghetti code, so testing for trans_a should be much easier now.

I've rebased the branch onto the current master, the tests are passing now.

src/generator_gemm_sse_avx_avx2_avx512.c
```diff
@@ -224,9 +235,125 @@ LIBXSMM_API_INTERN void libxsmm_generator_gemm_sse_avx_avx2_avx512_kernel( libxs
...
+    libxsmm_x86_instruction_push_reg( io_generated_code, LIBXSMM_X86_GP_REG_RDI );
+    libxsmm_x86_instruction_push_reg( io_generated_code, LIBXSMM_X86_GP_REG_R8 );
+    libxsmm_x86_instruction_push_reg( io_generated_code, LIBXSMM_X86_GP_REG_R9 );
+
+    /* the transpose microkernels called below use r8 for input and r9 for output so they are set here */
+    libxsmm_x86_instruction_alu_reg( io_generated_code, LIBXSMM_X86_INSTR_MOVQ, LIBXSMM_X86_GP_REG_RDI, LIBXSMM_X86_GP_REG_R8);
+    if ( ((LIBXSMM_GEMM_FLAG_USE_XGEMM_EXT_ABI & i_xgemm_desc->flags) != LIBXSMM_GEMM_FLAG_USE_XGEMM_EXT_ABI) ) {
+      libxsmm_x86_instruction_alu_reg( io_generated_code, LIBXSMM_X86_INSTR_MOVQ, l_transpose_stack_register, LIBXSMM_X86_GP_REG_R9);
+    } else {
+      libxsmm_generator_gemm_getval_stack_var( io_generated_code, &l_micro_kernel_config, LIBXSMM_GEMM_STACK_VAR_TRANSPOSE_PTR, LIBXSMM_X86_GP_REG_R9 );
```
kvoronin-intel, line 272: Here we set r9 (input for the meltwise = transpose) to the allocated space on the stack via the new API.

src/generator_gemm_sse_avx_avx2_avx512.c
```diff
@@ -476,38 +604,50 @@ LIBXSMM_API_INTERN void libxsmm_generator_gemm_sse_avx_avx2_avx512_kernel( libxs
...
   if ( ((LIBXSMM_GEMM_FLAG_USE_XGEMM_EXT_ABI & i_xgemm_desc->flags) == LIBXSMM_GEMM_FLAG_USE_XGEMM_EXT_ABI) ) {
     libxsmm_generator_gemm_destroy_stack_frame( io_generated_code, i_xgemm_desc, i_gp_reg_mapping, &l_micro_kernel_config );
   }
-}
+  else {
+    /* cleaning up the stack memory for the transpose */
+    if (i_xgemm_desc->flags & LIBXSMM_GEMM_FLAG_TRANS_A) {
 
+        /* cleaning up the space for transpose */
+        libxsmm_x86_instruction_alu_imm( io_generated_code, LIBXSMM_X86_INSTR_ADDQ, LIBXSMM_X86_GP_REG_RSP, l_trans_a_stack_size );
```
kvoronin-intel, line 642: Clean-up here is done also under condition that the new API is not useed.

kvoronin-intel:
@alheinecke, @egeor I've rebased the stack-transpose branch onto the latest enable_gemms_in_eqs branch. To me it all went smoothly, but please have a look (especially @egeor).

src/generator_gemm_common.c
```diff
@@ -1181,6 +1200,8 @@ int libxsmm_generator_gemm_get_rbp_relative_offset( libxsmm_gemm_stack_var stack
       return 72;
     case LIBXSMM_GEMM_STACK_VAR_ARG_10:
       return 80;
+    case LIBXSMM_GEMM_STACK_VAR_TRANSPOSE_PTR:
+      return -96;
```
kvoronin-intel, line 1204: Decided to introduce a new pointer for the stack structure rather than re-use some other buffer (like "eltwise buf1"). The rationale is that I'm afraid of nasty bugs which can appear in the future in case some new thing will need "eltwise buf1" for both eltwise and transpose.
If you're confident that this will never be the case and we can safely re-use one of the existing pointers instead of creating `LIBXSMM_GEMM_STACK_VAR_TRANSPOSE_PTR`, let me know.

egeor: Sounds good -- definitely this is better approach as we discussed.

src/generator_gemm_sse_avx_avx2_avx512.c
```diff
@@ -224,9 +235,125 @@ LIBXSMM_API_INTERN void libxsmm_generator_gemm_sse_avx_avx2_avx512_kernel( libxs
     libxsmm_generator_gemm_setup_stack_frame( io_generated_code, i_xgemm_desc, i_gp_reg_mapping, &l_micro_kernel_config);
   }
 
+  // Local variables used only for older gemm setup (not LIBXSMM_GEMM_FLAG_USE_XGEMM_EXT_ABI)
+  unsigned int l_trans_a_stack_size = 0;
+  unsigned int l_transpose_stack_register = LIBXSMM_X86_GP_REG_UNDEF; // for saving start of the A transpose allocated on the stack
+
+  if (i_xgemm_desc->flags & LIBXSMM_GEMM_FLAG_TRANS_A) {
+    if ( ((LIBXSMM_GEMM_FLAG_USE_XGEMM_EXT_ABI & i_xgemm_desc->flags) != LIBXSMM_GEMM_FLAG_USE_XGEMM_EXT_ABI) ) {
```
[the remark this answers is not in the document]
egeor: This should be fine since the stack manipulation infrastructure "requires" the new API

egeor:
@kvoronin-intel  This is an enhancement rather a change request:
Now the code allocates scratch in stack, transposes A there and uses the transposed A for the rest of the kernel.
This is all good, but with the new *ext API we support also generic ops on A/B before the xgemm:

https://github.com/hfp/libxsmm/blob/enable_gemm_in_eqs/include/libxsmm_typedefs.h#L919

So basically:
Ap = unary( A ) where the unary A may be transpose. In this case, if store_ap is set, the user provides a pointer where Ap should be saved (in this case the transposed A).

https://github.com/hfp/libxsmm/blob/enable_gemm_in_eqs/include/libxsmm_typedefs.h#L894

I believe with minimal changes you should be able to extend the current implementation as follows:
*  when the new *ext API is used and Ap = trans( A )  is requested you can use the equivalent logic as if LIBXSMM_GEMM_FLAG_TRANS_A was set (i.e. what you did) 

*  when the new *ext API is used and Ap = trans( A )  is requested, and *store_ap* is set, then instead of allocating scratch in stack, you should be able to use directly that user-provided pointer (e.g. by merely setting that proper ap pointer in the stack var LIBXSMM_GEMM_STACK_VAR_TRANSPOSE_PTR :) ). This is important in the context of building higher level contractions with this microkernel... Think of a large GEMM with transpose A. The user would block the loops and use the trans_A xgemm microkernel only the first time the "input/original A" panels are touched. In the rest iterations the user would use regular (non trans A) microkernels by reusing the transposed A from the provided buffer. This is crucial to get high performance A^T x B ops (since the transposed tensor will be reused/the transposed cost will be amortized).

Not sure if you want to do this in a different PR or update the current one.

alheinecke:
These suggestions are good, but orthogonal. A^T, B^T, C^T, A_VNNI, B_VNNI, C_VNNI need support without using the EXT API and can be implemented using EXT API as well. 
This PR enables some A^T on to of the B^T we already have. So in the end we will several changes to enable both cases and even combinations: In the EXT API we can have unary on A which is computing exp(A) and we can have the A^T flag being set.

src/generator_gemm.c
```diff
@@ -170,9 +170,28 @@ void libxsmm_generator_gemm_kernel( libxsmm_generated_code*        io_generated_
   }
 
   /* check LDA */
-  if ( l_xgemm_desc_mod.lda < l_xgemm_desc_mod.m ) {
-    LIBXSMM_HANDLE_ERROR( io_generated_code, LIBXSMM_ERR_LDA );
-    return;
+  if ( (l_xgemm_desc_mod.flags & LIBXSMM_GEMM_FLAG_TRANS_A) == LIBXSMM_GEMM_FLAG_TRANS_A ) {
+    // Neither non-FP32 nor BRGEMM are supported for the trans_a = 1 case
```
alheinecke, line 174: please use C89 comments ( /* */ ) instead of C++ comments ( // )

src/generator_gemm_common.c
```diff
@@ -804,10 +806,24 @@ void libxsmm_generator_gemm_setup_stack_frame_allocate_scratch( libxsmm_generate
     }
   }
 
+  /* Allocate scratch for stashing 32 zmms  */
+  if ( ((LIBXSMM_GEMM_FLAG_USE_XGEMM_EXT_ABI & i_xgemm_desc->flags) == LIBXSMM_GEMM_FLAG_USE_XGEMM_EXT_ABI) &&
```
alheinecke, line 810: Why is this guarded for EXT_ABI

src/generator_gemm_common.c
```diff
@@ -846,7 +862,8 @@ void libxsmm_generator_gemm_setup_stack_frame( libxsmm_generated_code*
    *      Eltwise output_ptr                        <-- RBP-64
    *      Eltwise buf1_ptr                          <-- RBP-72
    *      Eltwise buf2_ptr                          <-- RBP-80
-   *      Batch-reduce count                        <-- RBP-88, RSP
+   *      Batch-reduce count                        <-- RBP-88,
+   *      Transpose A ptr                           <-- RBP-96, RSP
```
alheinecke, line 866: it seems the list of variables in the comment is not aligned with the case statement, however, this is more issue in base and not in this PR (@egeor)

alheinecke (changes requested):

src/generator_gemm_common.c
kvoronin-intel: Good question. I've taken it from gemm scratch allocation a couple of lines above https://github.com/hfp/libxsmm/blob/enable_gemm_in_eqs/src/generator_gemm_common.c#L799

Main reason for doing so was that I was unsure whether we intend calling `libxsmm_generator_gemm_setup_stack_frame_allocate_scratch()` only when EXT_ABI is enabled or not.
On one hand, we are checking for this outside the call in https://github.com/hfp/libxsmm/blob/enable_gemm_in_eqs/src/generator_gemm_sse_avx_avx2_avx512.c#L214.
On the other hand, I saw this check in the function `libxsmm_generator_gemm_setup_stack_frame_allocate_scratch()` as well.

As long as we check in generator_gemm_sse_avx_avx2_avx512.c, I can definitely remove the check inside `..._allocate_scratch`.

egeor: The idea (before the A trans work) is that the the "zmm scratch" *would* be used only with the new EXT_ABI (since otherwise we wouldn't fuse anything in the first place). 

The libxsmm_generator_gemm_setup_stack_frame_allocate_scratch is intended to be called  either when we have SPR kernel (since we need that gemm scratch for dumping the tiles/downconvert) or when we have EXT_ABI (fusion in xgemm). This is why those conditions were there in the first place.

src/generator_gemm.c
kvoronin-intel: Fixed

src/generator_gemm_common.c
kvoronin-intel: Thanks for the explanation @egeor, I'm removing the check for the transpose scratch here then. Only the gemm kernel will stay responsible for checking for EXT_ABI (for the transpose).

src/generator_gemm_sse_avx_avx2_avx512.c
```diff
@@ -101,7 +101,6 @@ LIBXSMM_API_INTERN void libxsmm_generator_gemm_sse_avx_avx2_avx512_kernel_wrappe
   libxsmm_x86_instruction_close_stream_gemm( io_generated_code, &l_gp_reg_mapping, 0, i_xgemm_desc->prefetch );
```
alheinecke, line 101: Please add to the incl. all files that are need for function defintions:
libxsmm_generator_mateltwise_init_micro_kernel_config_fullvector
libxsmm_generator_meltw_setup_stack_frame
libxsmm_generator_transform_norm_to_normt_32bit_avx512_microkernel
libxsmm_generator_meltw_destroy_stack_frame

src/generator_gemm_sse_avx_avx2_avx512.c
```diff
@@ -224,9 +235,125 @@ LIBXSMM_API_INTERN void libxsmm_generator_gemm_sse_avx_avx2_avx512_kernel( libxs
...
+    libxsmm_mateltwise_gp_reg_mapping l_mateltwise_gp_reg_mapping;
+    memset(&l_mateltwise_gp_reg_mapping, 0, sizeof(l_mateltwise_gp_reg_mapping));
+#if defined(_WIN32) || defined(__CYGWIN__)
+    l_mateltwise_gp_reg_mapping.gp_reg_param_struct = LIBXSMM_X86_GP_REG_RCX;
+#else /* match calling convention on Linux */
+    l_mateltwise_gp_reg_mapping.gp_reg_param_struct = LIBXSMM_X86_GP_REG_RDI;
+#endif
+
+    /* stack management at the start for meltw kernel */
+    libxsmm_generator_meltw_setup_stack_frame( io_generated_code, l_mateltwise_desc, &l_mateltwise_gp_reg_mapping, &l_mateltwise_kernel_config);
```
alheinecke, line 312: We are setting up the GEMM stack and here also the meltw stack 
@egeor does this work in general or does it just work here as transpose is not using any stack variables...?

src/generator_gemm_sse_avx_avx2_avx512.c
```diff
@@ -224,9 +235,125 @@ LIBXSMM_API_INTERN void libxsmm_generator_gemm_sse_avx_avx2_avx512_kernel( libxs
...
+                                                                         l_gp_reg_in, l_gp_reg_out, l_gp_reg_mloop, l_gp_reg_nloop,
+                                                                        &l_mateltwise_kernel_config, l_mateltwise_desc );
+    } else if ( (io_generated_code->arch >= LIBXSMM_X86_GENERIC) && (io_generated_code->arch < LIBXSMM_X86_AVX) ) {
+        libxsmm_generator_transform_norm_to_normt_32bit_sse_microkernel( io_generated_code, io_loop_label_tracker,
+                                                                         l_gp_reg_in, l_gp_reg_out, l_gp_reg_mloop, l_gp_reg_nloop,
+                                                                        &l_mateltwise_kernel_config, l_mateltwise_desc );
+    }
+
+    /* stack management at the end for meltw kernel */
+    libxsmm_generator_meltw_destroy_stack_frame( io_generated_code, l_mateltwise_desc, &l_mateltwise_kernel_config );
```
alheinecke, line 332: see above the comment on creating the meltw stack

alheinecke (changes requested):
I checked out the PR and tested it. Please fix the following 3 merge blockers:
- and incl. for the meltw functions such that they are defined (GCC and clang won't even compile the code right now)
- please add support for FP64 transpose (otherwise "make test" fails)
- @egeor calling setting up the meltw stack-frame inside the GEMM kernel looks weird to me.

After these changes on x86 it should work. For aarch64 "make test" will still fail, but the entire target branch of this PR is not working well on aarch64...

src/generator_gemm_sse_avx_avx2_avx512.c
egeor: It does work in general... To put it differently, nested stack frames setup/destroy should work (and already tested in the mateq framework with the various fusions). That said, here nothing happens since no stack vars are used. I saw it already in the PR but didn't comment since to my eyes this is the right way to do it: if tomorrow for whatever reason we decide e.g. for performance to change the leveraged meltw TPP and make it use stack vars, then we don't have to change the GEMM code gen, or for that matter any other code gens that use the leveraged melts TPP. The setup/destroy stack frame will take care of the changes transparently.

egeor:
> I checked out the PR and tested it. Please fix the following 3 merge blockers:
> 
> * and incl. for the meltw functions such that they are defined (GCC and clang won't even compile the code right now)
> * please add support for FP64 transpose (otherwise "make test" fails)
> * @egeor calling setting up the meltw stack-frame inside the GEMM kernel looks weird to me.
> 
> After these changes on x86 it should work. For aarch64 "make test" will still fail, but the entire target branch of this PR is not working well on aarch64...



> I checked out the PR and tested it. Please fix the following 3 merge blockers:
> 
> * and incl. for the meltw functions such that they are defined (GCC and clang won't even compile the code right now)
> * please add support for FP64 transpose (otherwise "make test" fails)
> * @egeor calling setting up the meltw stack-frame inside the GEMM kernel looks weird to me.
> 
> After these changes on x86 it should work. For aarch64 "make test" will still fail, but the entire target branch of this PR is not working well on aarch64...

See my comment below regarding nested stack frames.

PR #576, opened 2022-01-20
Adding a new TPP sample for batch normalization

kvoronin-intel:
Summary:

- Refactored existing sample equation/equation_batchnorm.c along the lines of deeplearning/fullyconnected_tpp

<to be extended>


samples/deeplearning/common/dnn_common.h
```diff
@@ -2429,6 +2458,91 @@ LIBXSMM_INLINE void naive_fusedbatchnorm_fp(naive_fusedbatchnorm_t* param, const
   }
 }
 
+LIBXSMM_INLINE void naive_fusedbatchnorm_fp_fp64(naive_fusedbatchnorm_t* param, const double* input_ptr, double* output_ptr, const double* input_add_ptr,
```
kvoronin-intel, line 2461: Added naive implementations for fp64 in order to compute reference in fp64 for a more accurate error estimation

samples/deeplearning/fused_bn_gn_tpp/Makefile
```diff
@@ -0,0 +1,97 @@
...
+.PHONY: all
+all: $(XFILES)
+
+.PHONY: compile
+compile: $(OBJECTS) $(FTNOBJS)
+
+$(OUTDIR)/layer_example_f32: $(OUTDIR)/.make $(BLDDIR)/layer_example_f32-c.o $(LIBDEP) $(EXTDEP)
+	$(LD) -o $@ $(BLDDIR)/layer_example_f32-c.o $(call cleanld,$(EXTLIB) $(MAINLIB) $(SLDFLAGS) $(LDFLAGS) $(CLDFLAGS))
+
+#$(OUTDIR)/layer_example_bf16_amx: $(OUTDIR)/.make $(BLDDIR)/layer_example_bf16_amx-c.o $(LIBDEP) $(EXTDEP)
```
kvoronin-intel, line 61: To be removed...

samples/deeplearning/fused_bn_gn_tpp/layer_example_f32.c
```diff
@@ -0,0 +1,1190 @@
...
+  const libxsmm_blasint N  = cfg.N;
+  const libxsmm_blasint CP = cfg.CP;
+  const libxsmm_blasint HW = cfg.HW;
+  const libxsmm_blasint CB = cfg.CB;
+  const libxsmm_blasint num_HW_blocks = cfg.num_HW_blocks;
+
+  /* computing first logical thread */
+  const libxsmm_blasint ltid = my_tid - start_tid;
+
+  /* number of tasks that could be run in parallel for 1d blocking */
```
kvoronin-intel, line 177: Only 1d parallelization is implemented (no creation of row- and col-teams) at the moment. Same for bwd.

samples/deeplearning/fused_bn_gn_tpp/layer_example_f32.c
```diff
@@ -0,0 +1,1190 @@
...
+    (work_C / cfg.threads) : ((work_C / cfg.threads) + 1);
+  /* compute thr_begin and thr_end */
+  const libxsmm_blasint thr_begin_C = (ltid * chunksize_C < work_C) ? (ltid * chunksize_C) : work_C;
+  const libxsmm_blasint thr_end_C = ((ltid + 1) * chunksize_C < work_C) ? ((ltid + 1) * chunksize_C) : work_C;
+
+  /* lazy barrier init */
+  libxsmm_barrier_init(cfg.barrier, ltid);
+
+  /* FIXME: Set const modifiers properly? Cannot put const as then
+        reduce_HW_params.in.primary    = &LIBXSMM_VLA_ACCESS(4, inp, n, cp, hwb*(HW/num_HW_blocks), 0, CP, HW, CB); complains */
```
kvoronin-intel, line 201: Not sure what's the appropriate way to handle this const-issue. Ideally, const would be good to have here.

samples/deeplearning/fused_bn_gn_tpp/layer_example_f32.c
```diff
@@ -0,0 +1,1187 @@
...
+  libxsmm_matrix_eqn_function  dbeta_func;
+  libxsmm_matrix_eqn_function  din_func;
+  libxsmm_meltwfunction_unary  all_zero_kernel;
+  libxsmm_meltwfunction_binary add_kernel;
+  libxsmm_meltwfunction_unary  copy_kernel;
+  /*my_eltwise_fuse fuse_type;*/
+} my_bn_bwd_config;
+
+my_bn_fwd_config setup_my_bn_fwd(libxsmm_blasint N, libxsmm_blasint CP, libxsmm_blasint HW, libxsmm_blasint CB,
+                                 libxsmm_blasint num_HW_blocks, libxsmm_blasint threads ) {
```
kvoronin-intel, line 77: Didn't come up with a better name than the `num_HW_blocks` (taken from the existing sample in equation/)

samples/deeplearning/fused_bn_gn_tpp/layer_example_f32.c
```diff
@@ -0,0 +1,1187 @@
...
+  LIBXSMM_VLA_DECL(3, float, sum_X_X2, ((float*)scratch), CP, CB);  /* [2, CP, CB] */
+  LIBXSMM_ASSUME_ALIGNED(sum_X_X2_, 64);
+  const libxsmm_blasint sum_N_offset = (LIBXSMM_UP2((uintptr_t)(((float*)scratch) + CP * 2 * CB), 64) - ((uintptr_t)(scratch))) / sizeof(float);
+  LIBXSMM_VLA_DECL(3, float, sum_N, ((float*)scratch) + sum_N_offset, N, CB);  /* [CP, N, CB] */
+  LIBXSMM_ASSUME_ALIGNED(sum_N_, 64);
+  const libxsmm_blasint sumsq_N_offset = (LIBXSMM_UP2((uintptr_t)(((float*)scratch) + sum_N_offset + CP * N * CB), 64) - ((uintptr_t)(scratch))) / sizeof(float);
+  LIBXSMM_VLA_DECL(3, float, sumsq_N, ((float*)scratch) + sumsq_N_offset, N, CB);  /* [CP, N, CB] */
+  LIBXSMM_ASSUME_ALIGNED(sumsq_N_, 64);
+
+  { /* stupid block to keep indentation */
```
kvoronin-intel, line 216: Will be removed after the code review, leftover from older code

samples/deeplearning/fused_bn_gn_tpp/layer_example_f32.c
alheinecke: That's fine it's eltwise code.

samples/deeplearning/fused_bn_gn_tpp/layer_example_f32.c
alheinecke: We should make num_HW_blocks a parameter which is derived inside the setup, depending on size and we should also split HW into one parameter for H and for W. Also CP and CB is blocking/splitting inside of C, so we should provide C and the CB (the inner blocking, I hope this aligns with CNN)

alheinecke (changes requested):
see comments

samples/deeplearning/fused_bn_gn_tpp/layer_example_f32.c
kvoronin-intel: Implemented the changes (also renamed `CB` as `bc` following the naming in fullyconnected_tpp). For now, left num_HW_blocks as internal but hardcoded parameter, without any heuristic.

samples/deeplearning/fused_bn_gn_tpp/layer_example_f32.c
```diff
@@ -0,0 +1,1209 @@
...
+    for ( cpxnt = thr_begin_dN; cpxnt < thr_end_dN; ++cpxnt ) {
+      { /* stupid block to keep indentation */
+        n  = cpxnt%N;
+        cp = cpxnt/N;
+
+        libxsmm_matrix_arg arg_array[8];                                                               /* Private eqn args and params */
+        libxsmm_matrix_eqn_param eqn_param;
+        int hwb, cb;
+
+        /* FIXME: Replace expressions for pgamma, pdgamma etc. with ACCESS? */
```
kvoronin-intel, line 650: Do we want to put LIBXSMM_VLA_ACCESS here? For uniformity, it may make sense but the current form of the expressions in the for-loop looks too good to me to change it.

alheinecke: yes, let's put VLA everywhere

samples/deeplearning/fused_bn_gn_tpp/layer_example_f32.c
alheinecke: OK

samples/deeplearning/fused_bn_gn_tpp/layer_example_f32.c
```diff
@@ -0,0 +1,1209 @@
...
+#define COMPUTE_FP64_REFERENCE
+
+/* include c-based dnn library */
+#ifdef CNN_HEADER
+  #include "dnn_common.h"
+#else /* LIBXSMM sample */
+  #include "../common/dnn_common.h"
+#endif
+
+/* currently unused, as it is not clear that this will be used in CNN
```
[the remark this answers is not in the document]
alheinecke: you mean my_pass? It's for fullyconnected in BWD as we have two BWD passes (BWD_D and BWD_W) and we can run both, or just one of the them. For BN/ GN we only have BWD_D. Since we have FWD and BWD calls we don't need that here.

alheinecke:
should we merge in this state or are some more commits comming?

alheinecke:
BTW, while it's fine to skip padding/stride options. We need to support:

    if ( norm_type == 0 ) {
      if ( fuse_type == 0 ) {
        fusedbatchnorm_desc.fuse_ops = LIBXSMM_DNN_FUSEDBN_OPS_BN;
      } else if ( fuse_type == 1 ) {
        fusedbatchnorm_desc.fuse_ops = LIBXSMM_DNN_FUSEDBN_OPS_BN_RELU;
      } else if ( fuse_type == 2 ) {
        fusedbatchnorm_desc.fuse_ops = LIBXSMM_DNN_FUSEDBN_OPS_BN_ELTWISE;
      } else if ( fuse_type == 3 ) {
        fusedbatchnorm_desc.fuse_ops = LIBXSMM_DNN_FUSEDBN_OPS_BN_ELTWISE_RELU;
      } else if ( fuse_type == 4 ) {
        fusedbatchnorm_desc.fuse_ops = LIBXSMM_DNN_FUSEDBN_OPS_BN_RELU_WITH_MASK;
      } else if ( fuse_type == 5 ) {
        fusedbatchnorm_desc.fuse_ops = LIBXSMM_DNN_FUSEDBN_OPS_BN_ELTWISE_RELU_WITH_MASK;
      } else {
        /* shouldn't happen */
        return -1;
      }
    } 

as in the original driver.

kvoronin-intel:
Tue main goal of opening this PR was to gain general feedback. I think this is done so now I'm changing the state to WIP. Let's keep this PR pending (I will update it following the comments from @alheinecke and with changes implied by integration for CNN, which is in progress right now)

alheinecke:
changed the base to dnn_refactor
please split the code into driver and implementation, implementation goes into samples/op_lib_tpp

alheinecke:
you might want to merge dnn_refactor into your branch such that all the weird conflicts go away

alheinecke (commented):
while the batchnorm stuff is developing well, why are these 2/5 cnn files part of the PR?

samples/deeplearning/fused_bn_gn_tpp/layer_example_f32.c
kvoronin-intel: Done

samples/deeplearning/fusedbn_tpp/run_test.sh
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env bash
+set -eo pipefail
+
+# Source the compiler
+
+#export OMP_NUM_THREADS=1
+export CHECK=1
+
+iters=10
```
alheinecke, line 9: Can please add layers, e.g. copy from:
https://github.com/libxsmm/libxsmm/blob/master/samples/deeplearning/fusedbndriver/run_resnet50.sh#L99

alheinecke (changes requested):
very minor thing in the test script:
please copy over all resnet test cases. 

the code itself is ready for merge

samples/deeplearning/fusedbn_tpp/run_test.sh
kvoronin-intel: Is it still fine that the code doesn't support non-unit strides and non-zero padding (these cases are present in the resnet layers in the script)?

alheinecke: yes, pytroch doesn't support physically padded tensor easily and the stride was to fuse subsampling, but we do this in pytorch in the convolutions,

PR #583, opened 2022-02-03
A fix and minor improvements for batchnorm TPP

kvoronin-intel:
Summary:

- A fix for the hardcoded HW block size (now it's max(H,W)).
- Renaming batchnorm fuse_type enum to avoid potential clashes with groupnorm TPP.
- Moved ReLU+mask/eltwise fusion from separate loop into the existing one (for dgamma/dbeta), should be better performing and looks functionally correct.


samples/deeplearning/op_lib_tpp/batchnorm_tpp.h
```diff
@@ -243,12 +241,12 @@ my_bn_fwd_config setup_my_bn_fwd(libxsmm_blasint N, libxsmm_blasint C, libxsmm_b
 
   my_eqn10 = libxsmm_matrix_eqn_create();                          /* y = (s*x + b)*gamma + beta */
 
-  ternary_flags               = LIBXSMM_MELTW_FLAG_TERNARY_BCAST_COL_IN_1 | LIBXSMM_MELTW_FLAG_TERNARY_BCAST_COL_IN_2 | LIBXSMM_MELTW_FLAG_TERNARY_REUSE_IN_2_AS_OUT;
+  ternary_flags               = (libxsmm_meltw_ternary_flags)(LIBXSMM_MELTW_FLAG_TERNARY_BCAST_COL_IN_1 | LIBXSMM_MELTW_FLAG_TERNARY_BCAST_COL_IN_2 | LIBXSMM_MELTW_FLAG_TERNARY_REUSE_IN_2_AS_OUT);
   op_metadata[0].eqn_idx      = my_eqn10;
```
alheinecke, line 245: ternary_flags should be of type libxsmm_bitfield, so the cast to libxsmm_meltw_ternary_flags should not be needed.
https://github.com/libxsmm/libxsmm/blob/master/include/libxsmm.h#L579

This is try for all these similar changes in this PR.

alheinecke (changes requested):
I don't understand the cast, may be this cross-over as we are still patching the frontend?!

samples/deeplearning/op_lib_tpp/batchnorm_tpp.h
kvoronin-intel: The cast was inspired by the CNN integration where (at least an older GCC 8.3.0) complained about this. I guess the logical operations were done with a conversion to integer and an implicit cast back to the enum was not good enough for GCC.

alheinecke: ah I see, yes GCC complains about, I guess "ternary_flags" and all the others are not of type libxsmm_bitfield as this recently changed as other compilers complained that the ORed value is not defined in the enum. In order to avoid all of this we have now libxsmm_bitfield (which is a simply typedef for now to unsigned int) where we can run in future tests if we want.

kvoronin-intel: Right, now I changed the type to libxsmm_bitfield.

samples/deeplearning/op_lib_tpp/batchnorm_tpp.h
```diff
@@ -1036,7 +936,23 @@ void my_bn_bwd_exec( my_bn_bwd_config cfg, float *pdout, const float *pinp, cons
     arg_array[6].primary = (void*)&LIBXSMM_VLA_ACCESS(2, gamma, cp, 0, bc);
 
     for(hwb=0; hwb < num_HW_blocks; hwb++){
-
+      if (cfg.fuse_type == MY_BN_FUSE_ELTWISE ||
```
kvoronin-intel, line 939: Moving the ReLU and eltwise here should be alright as same blocks are being modified which are used for dgamma, dbeta computations by thee same threads.

alheinecke (approved):

PR #589, opened 2022-02-15
Enabling transpose of A in GEMM via an intermediate transpose on the stack (revived)

kvoronin-intel:
Previous PR:
https://github.com/libxsmm/libxsmm/pull/564#pullrequestreview-860357927

Last comments there:

- add incl. for the meltw functions such that they are defined (GCC and clang won't even compile the code right now)
Done?
- please add support for FP64 transpose (otherwise "make test" fails)
Done?
- @egeor calling setting up the meltw stack-frame inside the GEMM kernel looks weird to me.
Deferred by @egeor? 

The issues will be addressed in this PR.

Diff is looking weird, hopefully it's a github issue with matching revisions for the preliminary diff.


egeor:
@kvoronin-intel calling setting up the meltw stack-frame inside the GEMM kernel looks weird to me

I thought I had commented on this. In any case, in this specific case it is not really needed but calling it is the right way to go forward it in my eyes... the stack-frame infrastructure is composable. In this case nothing happens for this meltw op, but if we change the JITer of the meltw op tomorrow to leverage stack (e.g. because we find a more performant JIT variant) then you'd have to add back these stack-frame related calls (which will be properly handling the stack-frame)...

alheinecke (changes requested):
Can you please also add test scripts for F64?

Was the code tested for F32/FP64 with all x86 targets? 

I will do the following create and branch and the setup CI with the test scripts enabled. We can then also add aarch64 support and then merge it back.

kvoronin-intel:
> Can you please also add test scripts for F64?
> 
> Was the code tested for F32/FP64 with all x86 targets?
> 
> I will do the following create and branch and the setup CI with the test scripts enabled. We can then also add aarch64 support and then merge it back.

Added FP64 test scripts. Even for kernel_fused_ops though it does not support fp64 (is this expected or should attempt to enable it?).
Tested manually on avx512 and avx2 machines but would rather do a broader automated testing if possible.

PR #591, opened 2022-02-17
Handle errors in the PyFR sample.

FreddieWitherden:
If the generated kernel gives different results to the sparse kernel, the driver now returns 1.  This will be used in the future for unit testing of the routines.


alheinecke (commented):
Can you please also amend the script to cover more test cases?

FreddieWitherden:
Updated.  When running the examples I get a few failures.  This is not entirely unexpected (some of the kernels are larger than the default code buffer size).  What is perhaps more interesting is that we fail with a stack smashing attempt which implies somewhere we're not being as careful as we should be (I am unsure if this is in the sample code or libxsmm itself, if it is the latter we should probably fix the issue ASAP).

alheinecke (approved):

Issue #592, opened 2022-02-21
Malloc errors on Apple M1


WillTrojak:
I have been having some trouble getting libxsmm to work on the Apple M1 architecture 
```bash
$ sysctl -a | grep brand
> machdep.cpu.brand_string: Apple M1
```
 
I compiled the latest master branch (67e5663) with
```bash
$ make STATIC=1 CXX=g++ CC=gcc FC=
```
I then compiled and downloaded the PyFR test. Running `/libxsmm/samples/pyfr/test.sh` I get the following error:

```bash
$ . ./test.sh 
pyfr_driver_asp_reg(78775,0x1048c0580) malloc: *** error for object 0x1048fc000: pointer being freed was not allocated
pyfr_driver_asp_reg(78775,0x1048c0580) malloc: *** set a breakpoint in malloc_error_break to debug
[3]    78775 abort      ./pyfr_driver_asp_reg $m 48000 10 > /dev/null
./mats/p1/hex/m0-sp.mtx failed!
pyfr_driver_asp_reg(78776,0x1049b8580) malloc: *** error for object 0x104a38000: pointer being freed was not allocated
pyfr_driver_asp_reg(78776,0x1049b8580) malloc: *** set a breakpoint in malloc_error_break to debug
[3]    78776 abort      ./pyfr_driver_asp_reg $m 48000 10 > /dev/null
./mats/p1/hex/m132-sp.mtx failed!
...
```

Recompiling with `$ make STATIC=0 CXX=g++ CC=gcc FC=` and trying to actually run PyFR 1.13.0 I get the following:

```bash
Python(65961,0x1030e0580) malloc: *** error for object 0x1172e8000: pointer being freed was not allocated
Python(65961,0x1030e0580) malloc: *** set a breakpoint in malloc_error_break to debug
[Wills-MBP:65961] *** Process received signal ***
[Wills-MBP:65961] Signal: Abort trap: 6 (6)
[Wills-MBP:65961] Signal code:  (0)
[Wills-MBP:65961] [ 0] 0   libsystem_platform.dylib            0x000000019c6904e4 _sigtramp + 56
[Wills-MBP:65961] [ 1] 0   libsystem_pthread.dylib             0x000000019c678eb0 pthread_kill + 288
[Wills-MBP:65961] [ 2] 0   libsystem_c.dylib                   0x000000019c5b6314 abort + 164
[...]

PR #594, opened 2022-02-25
Do not attempt to free JIT kernels on AARCH64 macOS.

FreddieWitherden:


FreddieWitherden:
We'll need to uncomment these lines when a proper fix is available.  For PyFR the fact we'll leak a few megabytes of memory really is not an issue, but it would still be nice to have it fixed.

alheinecke (approved):

alheinecke:
yes and that's why I leave the issue #592 open for now. This was just a suggestion to get you guys unblocked on M1.

PR #605, opened 2022-03-22
[batchnorm TPP] Adding scale-only norm type support for batchnorm TPP (and the test driver)

kvoronin-intel:
Summary:

- Adding `norm_type` to batchnorm TPP and the driver in samples/deeplearning/fusedbn_tpp
- Renaming enum for fuse type to be consistent with the groupnorm
- Minor changes in the driver

Tested locally with `norm_type = 0, 1`, both work (compared to naive implementation in dnn_common.h).

### DIFF
--- a/samples/deeplearning/op_lib_tpp/batchnorm_tpp.h
+++ b/samples/deeplearning/op_lib_tpp/batchnorm_tpp.h
@@ -875,29 +875,10 @@ void my_bn_bwd_exec( my_bn_bwd_config cfg, float *pdout, const float *pinp, cons
   float alpha = 0.0f;
   LIBXSMM_VLA_DECL(4, const unsigned char, relumask, prelumask, CP, HW, bc/BITS_PER_CHAR);    /* [N, CP, HW, bc/BITS_PER_CHAR] */
 
-  const libxsmm_blasint dbeta_N_offset = (LIBXSMM_UP2((uintptr_t)(((float*)scratch) + CP * N * bc), 64) - ((uintptr_t)(scratch))) / sizeof(float);
-  LIBXSMM_VLA_DECL(3, float, dgamma_N, ((float*)scratch),                  N, bc);  /* [CP, N, bc] */
-  LIBXSMM_ASSUME_ALIGNED(dgamma_N_, 64);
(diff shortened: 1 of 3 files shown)


samples/deeplearning/op_lib_tpp/batchnorm_tpp.h
```diff
@@ -21,14 +21,19 @@
...
   MY_BN_FUSE_ELTWISE = 2,
   MY_BN_FUSE_ELTWISE_RELU = 3,
   MY_BN_FUSE_RELU_WITH_MASK = 4,
   MY_BN_FUSE_ELTWISE_RELU_WITH_MASK = 5
-} my_normalization_fuse;
+} my_bn_fuse;
+
+typedef enum my_bn_norm_type {
+  MY_BN_FULL_NORM  = 0, /* stats + normalize for fwd, all grads for bwd */
+  MY_BN_SCALE_ONLY = 1  /* normalize only for fwd, only input grad for bwd */
```
kvoronin-intel, line 35: Not 100% sure if it makes more sense to separate fwd/bwd behavior with different enum values or not. Currently, combined them as is done in the naive implementation in dnn_common.h. Since it is a runtime parameter for fwd/bwd_exec

samples/deeplearning/op_lib_tpp/batchnorm_tpp.h
```diff
@@ -891,130 +902,133 @@ void my_bn_bwd_exec( my_bn_bwd_config cfg, float *pdout, const float *pinp, cons
...
 
-  for ( cpxnt = thr_begin_dN; cpxnt < thr_end_dN; ++cpxnt ) {
-    n  = cpxnt%N;
-    cp = cpxnt/N;
-
-    int hwb, cb;
-
-    LIBXSMM_ALIGNED(float lcl_dgamma_ptr[bc], 64);
-    LIBXSMM_ALIGNED(float lcl_dbeta_ptr[bc], 64);
+  if (norm_type == MY_BN_FULL_NORM) {
```
kvoronin-intel, line 890: Only if-condition is being added, the actual code remains the same (except moving some variables declaration and initialization inside the if-case if they are only needed for the full norm)

alheinecke (approved):

PR #608, opened 2022-03-30
SR parameterized test, BF16 conv fix

nkmellem:


samples/eltwise/run_eltwise_avx512.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
```
alheinecke, line 1: please remove the entire script

samples/eltwise/eltwise_common.h
```diff
@@ -14,6 +14,118 @@
 #include <math.h>
 #include <string.h>
```
alheinecke, line 16: can you please remove this code

include/libxsmm_typedefs.h
```diff
@@ -337,7 +337,7 @@ typedef enum libxsmm_meltw_unary_type {
   LIBXSMM_MELTW_TYPE_UNARY_DECOMPRESS_SPARSE_FACTOR_16  = 49,
   LIBXSMM_MELTW_TYPE_UNARY_DECOMPRESS_SPARSE_FACTOR_32  = 50,
   LIBXSMM_MELTW_TYPE_UNARY_GATHER                       = 51,
-  LIBXSMM_MELTW_TYPE_UNARY_SCATTER                      = 52
+  LIBXSMM_MELTW_TYPE_UNARY_SCATTER                      = 52,
```
alheinecke, line 340: can you please remove the comma

samples/eltwise/run_test_avx512.sh
```diff
@@ -0,0 +1,99 @@
+#!/usr/bin/env bash
```
alheinecke, line 1: why this new script? For full support (=avx512) we have run_test.sh, we just have AVX2/SSE/aarch64 scripts as they don't implement all TPPs right now. Therefore I suggest removing this script.

src/generator_common_x86.c
```diff
@@ -1960,21 +1960,23 @@ void libxsmm_generator_vcvtneps2bf8_generic_avx512_prep_stack( libxsmm_generated
...
 
-
 LIBXSMM_API_INTERN
 void libxsmm_generator_vcvtneps2bf16_avx512_clean_stack( libxsmm_generated_code* io_generated_code,
                                                          const unsigned int      io_gp_reg ) {
   libxsmm_x86_instruction_pop_reg( io_generated_code, io_gp_reg );
   libxsmm_x86_instruction_pop_reg( io_generated_code, io_gp_reg );
   libxsmm_x86_instruction_pop_reg( io_generated_code, io_gp_reg );
   libxsmm_x86_instruction_pop_reg( io_generated_code, io_gp_reg );
-  libxsmm_x86_instruction_pop_reg( io_generated_code, io_gp_reg );
```
alheinecke, line 1971: Why is touching this routine needed this is also used in GEMM... is this because originally we had only 4 pops?

alheinecke (changes requested):

src/generator_common_x86.c
nkmellem: yes, it originally had only 4 POPs, i was trying to reuse this code which introduced that bug causing BF16 kernels to fail. its restored now.

samples/eltwise/run_eltwise_avx512.sh
nkmellem: Done.

samples/eltwise/eltwise_common.h
nkmellem: Done.

include/libxsmm_typedefs.h
nkmellem: done.

alheinecke (approved):

PR #610, opened 2022-04-01
BF8 support for eltwise kernels and GEMM

alheinecke:


alheinecke (commented):
serveral opens as highlighted before, let's review again, when the next set of commits is in,

alheinecke (commented):
just waiting for AVX512_VL256 reductions and then this is ready for merge to master

PR #615, opened 2022-04-06
AARCH64 SVE eltwise support

alheinecke:


src/generator_common_aarch64.c
```diff
@@ -1144,17 +1135,8 @@ void libxsmm_generator_prepare_coeffs_gelu_inv_ps_minimax3_aarch64_sve( libxsmm_
                                                                         const unsigned int                    i_vec_shifter,
                                                                         const unsigned int                    i_vec_half,
                                                                         const unsigned int                    i_vec_c0,
-                                                                        const unsigned int                    i_vec_c01,
-                                                                        const unsigned int                    i_vec_c02,
-                                                                        const unsigned int                    i_vec_c03,
```
AntonioNoack: These will be needed again, when GELU is implemented for 128/256 bit SVE.

AntonioNoack:
I found the fix for Dropout, will push it (or similar) soon :)
It was a bug, that I had fixed in the past, but replaced when I removed my old performance tests.

The issue is/was that the dropout test is using it's own function to determine the vector length, and it "detects" 4 for SVE (which is ofc incorrect).

alheinecke:
@AntonioNoack can you please re-run this branch's latest commit on A64FX ( https://github.com/libxsmm/libxsmm/blob/aarch64_eltwise/samples/eltwise/run_test_aarch64_sve.sh ) . I thought relu, dropout, gelu should work and transpose should simply fall back to the ASIMD code. It's important that we can run run_test_aarch64.sh without comments, so the best is to have all the transforms just falling back to the ASIMD generators. When this is all confirmed we are ready for prime time and can merge to master.

alheinecke:
@AntonioNoack @breuera : I have updated https://github.com/libxsmm/libxsmm/blob/aarch64_eltwise/samples/eltwise/run_test_aarch64_sve.sh  and ran it on A64FX. However, relu and dropout show random seg faults, but I saw in some report performance numbers for those. Can you please fix such that we can merge?

alheinecke:
right now the failing tests are commented in https://github.com/libxsmm/libxsmm/blob/aarch64_eltwise/samples/eltwise/run_test_aarch64_sve.sh

AntonioNoack:
I've found one solution until now :).
It solves the backwards passes for dropout & relu (my code was writing too many bytes for the masks).
Forward still has issues.

alheinecke:
Great, thanks for the update

PR #620, opened 2022-04-08
Improve SVE code generation in asparse_reg.

FreddieWitherden:
Here we play some neat tricks to improve the kernels.  Firstly, in the constants loaded from memory mode, multiple GPR's are pre-displaced enabling constants to be broadcast from memory with just a single load instruction (GPR + imm) with the spacing between registers being the displacement limit of the immediate.  Secondly, we now always pack constants into registers, allocating z0 if needed to serve as a scratch register (as SVE's indexed FMA is not orthogonal and only supports registers up to z7/z15) to mov higher number registers into.  On any implementation with mov elimination this should be very cheap and eliminated at the frontend.


alheinecke (commented):
The LIBXSMM_AARCH64_INSTR_SVE_ORR_V instruction is added also in the aarch64_eltwise branch but with a different encoding and the also other encodings change. 

We are right now in the process of merging aarch64_eltwise into master (within the next couple of days, as early as today). I suggest re-basing this PR to aarch64_eltwise (when this branch is still there when you have the chance) or then to the new master.

FreddieWitherden:
Okay, as there is no rush I'll wait until the element wise work is merged into master and then rebase.

alheinecke:
the aarch64_eltwise branch just landed. Can you please resolve the conflicts?

FreddieWitherden:
Done.

alheinecke (approved):

PR #621, opened 2022-04-09
Fixes for test script eltwise commit

deeptiag1:
Failing test:
./kernel_test/reduce_add_cols_x_x2_16b_gtld.sh
./kernel_test/reduce_add_cols_x_x2_32b_eqld.sh
./kernel_test/reduce_add_cols_x_x2_32b_gtld.sh
./kernel_test/reduce_add_rows_x_16b_eqld.sh
./kernel_test/reduce_add_rows_x_16b_gtld.sh
./kernel_test/reduce_add_rows_x2_16b_eqld.sh
./kernel_test/reduce_add_rows_x2_16b_gtld.sh
./kernel_test/reduce_add_rows_x2_32b_eqld.sh
./kernel_test/reduce_add_rows_x2_32b_gtld.sh
./kernel_test/reduce_add_rows_x_32b_eqld.sh
./kernel_test/reduce_add_rows_x_32b_gtld.sh
./kernel_test/reduce_add_rows_x_x2_16b_eqld.sh
./kernel_test/reduce_add_rows_x_x2_16b_gtld.sh
./kernel_test/reduce_add_rows_x_x2_32b_eqld.sh
./kernel_test/reduce_add_rows_x_x2_32b_gtld.sh
./kernel_test/reduce_max_rows_16b_eqld.sh
./kernel_test/reduce_max_rows_16b_gtld.sh
./kernel_test/reduce_max_rows_32b_eqld.sh
./kernel_test/reduce_max_rows_32b_gtld.sh

Debugging it

PR #624, opened 2022-04-11
Avx512 Binary fix

deeptiag1:


alheinecke (changes requested):
Seems this fix is only half-way correct... as this is missing:
https://github.com/libxsmm/libxsmm/blob/master/src/generator_mateltwise_sse_avx_avx512.c#L391

It's better to avoid all of the code duplication, for one setting of VL...

alheinecke (commented):
Is this now fixing all problems as this file was *not* updated:
https://github.com/libxsmm/libxsmm/blob/avx512vl/samples/eltwise/run_test_avx512_vl256.sh

alheinecke (commented):
When do we expect the missing test to be passing?

src/generator_mateltwise_reduce_avx_avx512.c
```diff
@@ -1133,12 +1133,23 @@ void libxsmm_generator_reduce_rows_avx512_microkernel( libxsmm_generated_code*
       libxsmm_generator_load_vreg_minus_infinity(io_generated_code, 'y', LIBXSMM_X86_GP_REG_RAX, aux_vreg);
     }
 
-    libxsmm_generator_mateltwise_initialize_avx_mask(io_generated_code, mask_out, 1);
-    /* Calculate input mask in case we see m_masking */
-    if (use_m_masking == 1) {
-      mask_reg = available_vregs-1;
-      libxsmm_generator_mateltwise_initialize_avx_mask(io_generated_code, mask_reg, m % vlen);
-      available_vregs--;
+    if (io_generated_code->arch <= LIBXSMM_X86_AVX2) {
```
alheinecke, line 1136: We are we following the AVX2 code path and add logic here instead of using the AVX512 path?

alheinecke (commented):
I'm not sure if I understand why the AVX2 code patch changes and not the AVX512 for reduce

src/generator_mateltwise_reduce_avx_avx512.c
deeptiag1: Yes, AVX512 code path had a lot of hard coded variable, hence we ended up using the AVX2 code path.
For BF16 we have two options:
1.  AVX2 code path doesn't support BF16 so we need to add logic. 
2. Rewrite AVX512 code path in a flexible way that it can be extended to AVX512_VL ISA

alheinecke: Please go with 2. and revert the changes to the AVX2 code. This is much cleaner and is much faster as this just halving the shuffle network. Touching up the AVX2 code leads to unmaintainable spaghetti code....

alheinecke: Also the changes to libxsmm_generator_mn_code_block_replicate_col_var_avx_avx512 and libxsmm_generator_replicate_col_var_avx_avx512_microkernel didn't look correct in #621

samples/eltwise/run_test_avx512_vl256.sh
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-export LIBXSMM_TARGET=avx512_vl256_clx
+# export LIBXSMM_TARGET=avx512_vl256_clx
```
alheinecke, line 3: This makes no sense...

src/generator_mateltwise_reduce_avx_avx512.c
```diff
@@ -462,7 +474,7 @@ void libxsmm_generator_reduce_cols_avx512_microkernel( libxsmm_generated_code*
     return;
   }
 
-  if (io_generated_code->arch < LIBXSMM_X86_AVX512) {
+  if (io_generated_code->arch < LIBXSMM_X86_AVX512_VL256) {
     vlen = 8;
     tmp_vreg = 15;
     max_m_unrolling = 15;
```
alheinecke, line 480: In line 553 there is check for CPX which doesn't seem to be handle in AVX512_VL256... don't know why

alheinecke (changes requested):
Started to review, but stopped after two things as I believe the code is not yet ready for review

src/generator_mateltwise_reduce_avx_avx512.c
alheinecke: On top the file conflicting with upstream an the PR cannot be merged

deeptiag1: yes, Fixing the conflict.

deeptiag1:
Closing this one and creating a cleaner Pull request:
https://github.com/libxsmm/libxsmm/pull/633/files

I have addressed the CPX comment [https://github.com/libxsmm/libxsmm/pull/633/files](url)
The unprompted func is libxsmm_generator_reduce_cols_ncnc_avx512_microkernel

PR #626, opened 2022-04-12
Unifying batchnorm TPP drivers for bf16 and fp32 with a single test

kvoronin-intel:
Description:
Same change as was done for the groupnorm, combining fp32 abd bf16 drivers for the TPP code in a single source file.


alheinecke (commented):
Why do we have to run_resnet50 files?

kvoronin-intel:
> Why do we have to run_resnet50 files?

One is inherited from the handle-based sample (deeplearning/fusedbndriver) but I consider it out-dated and having incorrect padding settings and hence created run_resnet50_cnn_opt.sh. I'll double check which one makes more sense and leave only one.

kvoronin-intel:
> > Why do we have to run_resnet50 files?
> 
> One is inherited from the handle-based sample (deeplearning/fusedbndriver) but I consider it out-dated and having incorrect padding settings and hence created run_resnet50_cnn_opt.sh. I'll double check which one makes more sense and leave only one.

Now there is only one left, @alheinecke. Let me know if you have other comments to address.

alheinecke (approved):

PR #629, opened 2022-04-13
Removed the redundant and too strict a check for datatypes for the gemm generator

kvoronin-intel:
Summary:
Optional creation of a modified gemm descriptor (due to the optional presence of A transpose flag) had an extra check for the datatype and disallowed bf16. Now this extra check is removed.

### DIFF
--- a/src/generator_gemm_sse_avx_avx2_avx512.c
+++ b/src/generator_gemm_sse_avx_avx2_avx512.c
@@ -149,28 +149,14 @@ LIBXSMM_API_INTERN void libxsmm_generator_gemm_sse_avx_avx2_avx512_kernel( libxs
 
   /* in case when A needs to be transposed, we need to change temporarily the desciptor dimensions for gemm */
   lda_transpose = i_xgemm_desc->m;
-  if (LIBXSMM_DATATYPE_F32 == (libxsmm_datatype)(i_xgemm_desc->datatype)) {
-    l_xgemm_desc_opa = (i_xgemm_desc->flags & LIBXSMM_GEMM_FLAG_TRANS_A ?
-      libxsmm_sgemm_descriptor_init(&l_blob_opa, i_xgemm_desc->m, i_xgemm_desc->n, i_xgemm_desc->k,
-        lda_transpose,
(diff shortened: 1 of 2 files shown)


src/generator_gemm_sse_avx_avx2_avx512.c
```diff
@@ -149,28 +149,14 @@ LIBXSMM_API_INTERN void libxsmm_generator_gemm_sse_avx_avx2_avx512_kernel( libxs
...
-        i_xgemm_desc->ldc,
-        1. /*alpha* is unused but used in the BYPASS check? */, (i_xgemm_desc->flags & LIBXSMM_GEMM_FLAG_BETA_0 ? 0.0 : 1.0) /*beta is already in the flags?*/,
-        (unsigned int)((unsigned int)(i_xgemm_desc->flags) & (~LIBXSMM_GEMM_FLAG_TRANS_A)), i_xgemm_desc->prefetch)
-      : i_xgemm_desc);
-  } else {
-    LIBXSMM_HANDLE_ERROR( io_generated_code, LIBXSMM_ERR_UNSUP_DATATYPE );
-    return;
-  }
+  l_xgemm_desc_opa = (i_xgemm_desc->flags & LIBXSMM_GEMM_FLAG_TRANS_A ?
+    libxsmm_sgemm_descriptor_init(&l_blob_opa, i_xgemm_desc->m, i_xgemm_desc->n, i_xgemm_desc->k,
```
alheinecke, line 153: Shouldn't there be a F32 check when calling sgemm_.....? I think we still need to check that Trans_A is only valid for FP32 and FP64 and not BRGEMM.

alheinecke (changes requested):
I think we to check for A_TRANS only for F32/F64

src/generator_gemm_sse_avx_avx2_avx512.c
kvoronin-intel: Sorry for the lame fix. You're right, the check still needs to be in-place. I've put it back but modified to avoid the error exit for valid cases.

PR #635, opened 2022-04-20
Add c brgemm prefetch

alheinecke:


src/generator_gemm_amx_microkernel.c
```diff
@@ -902,6 +919,10 @@ void libxsmm_generator_gemm_amx_microkernel( libxsmm_generated_code*
   int n_CL_to_pf;
   unsigned int tile_compute_instr = 0;
   unsigned int gp_reg_a;
+  unsigned int prefetch_C_scratch   = atoi(getenv("C_SCR_PF"));
```
alheinecke, line 922: before merging into master, we need default value as the current code will create issues.... can you pleas add

PR #638, opened 2022-05-04
Gather Scatter fix for AVX512VL

deeptiag1:


alheinecke:
This PR seems to be not tested against upstream (which would not care about this PR as AVX512_VL256 at a different place is redirected to AVX2), closing as we will fix it manually.

PR #653, opened 2022-05-26
aarch64 SVE256 support

alheinecke:
In order to support AWS Graviton3 (C7G) instances this PR aims at extending A64FX to SVE256


FreddieWitherden:
Lines which may need inspection:

 - https://github.com/libxsmm/libxsmm/blob/master/src/libxsmm_fsspmdm.c#L35 (should detect NEON without SVE)
 - https://github.com/libxsmm/libxsmm/blob/master/src/generator_spgemm.c#L127 (should detect SVE)

alheinecke:
yes the asparse reg kernel hasn't been touched so far. We are just gathering here changes to kernels and have created the sve256_support branch. If you have time to push fixes against that branch, that's highly appreciated.

Re: cpuid -> yeah that needs to be revisited.... The current solution segfaults and is therefore commented.

FreddieWitherden:
I think this:

https://developer.arm.com/documentation/ddi0595/2020-12/AArch64-Registers/ID-AA64PFR0-EL1--AArch64-Processor-Feature-Register-0

should enable us to detect SVE support.  The length is easy enough to get via `addvl`.

PR #661, opened 2022-06-17
Buffer size checks for AARCH64.

FreddieWitherden:
Probably worth running through the tests before merging (there are a bunch of generating functions my code path do not normally call).


hfp:
Thank you!

PR #662, opened 2022-06-23
Updated batchnorm fwd to optionally use new reduce TPP with beta = 0 and 1 (fp32 and bf16)

kvoronin-intel:
Summary:

- Updated batchnorm forward to use the newly added reduce TPP with beta = 0 and beta = 1 instead of explicitly zeroing out + reduction + addition TPPs (both fp32 and bf16). As a consequence, scratch is used differently. Previously it had two arrays for sums and sums of squares, each of size [CB, N, bc] but now it's a single array of double sizes which has sums and sums of squares intermingled (so it's of size [CB, N, 2*bc].

Verified locally on a single size that it is functionally correct.


samples/deeplearning/libxsmm_dnn/src/libxsmm_dnn_fusedbn.c
```diff
@@ -731,26 +759,86 @@ LIBXSMM_API void libxsmm_dnn_bn_fwd_exec_f32( libxsmm_dnn_bn_fwd_config cfg, con
...
       /*   sumsq_ncp_ptr[cb] = 0.0f;  */
       /* } */
 
-      LIBXSMM_ALIGNED(float lcl_sum_X_X2[2*bc], 64);
       if (cfg.use_hw_blocking == 0) { /* w-blocking */
-        reduce_param.out.primary = lcl_sum_X_X2;                                                    /* [2*bc]  */
-        for (hi = 0; hi < H; hi++) {
-          for (wb = 0; wb < num_W_blocks; wb++) {
-            reduce_param.in.primary = (void*)&LIBXSMM_VLA_ACCESS(5, inp, n, cp, hi, wb*(W/num_W_blocks), 0, CP, ifhp, ifwp, bc);
+        if (cfg.use_new_reduce) {
```
kvoronin-intel, line 788: @egeor, 
1) any comments about this new code / did you mean some other way of using new reduce for computing batchnorm stats?
2) do you think we should keep the old way (case cfg.use_new_reduce = 0) or should the new one be unconditionally on? (I'll do similar change for the extension code)

alheinecke (approved):

PR #663, opened 2022-06-23
Minor tweaks to the AARCH64 NEON path in asparse_reg.

FreddieWitherden:
Some ARM chips do not understand XOR as a zeroing idiom.  This avoids zeroing registers, avoiding the issue.


hfp:
Code now fails for ARM like on Jetson AGX (`LIBXSMM_AARCH64_INSTR_ASIMD_FNEG_V` -> `0x2ea0f80a`):

```
libxsmm_aarch64_instruction_asimd_compute: got 3 registers, but instruction has only 2: 0x2ea0f80a
```

PR #675, opened 2022-07-24
aarch64 SVE support (limited to A64FX) for dense and sparse Tensor Multiplications

alheinecke:


alheinecke (commented):
all EDGE unit tests (samples/edge) passed on Isambard

PR #676, opened 2022-07-25
Fixed missing init of SVE predicate registers when storing TPP register blocks.

breuera:
Most element-wise kernels have the predicate register still set from the loads when executing the stores.
XOR doesn't load and set them (especially p1); this fixes this with the issue of setting them multiple times (see code comment).

### DIFF
--- a/src/generator_mateltwise_unary_binary_aarch64.c
+++ b/src/generator_mateltwise_unary_binary_aarch64.c
@@ -505,6 +505,8 @@ void libxsmm_store_aarch64_2d_reg_block( libxsmm_generated_code*                
   unsigned int l_ld_bytes = i_mateltwise_desc->ldo * i_micro_kernel_config->datatype_size_out;
   unsigned int l_m_adjust = ( i_mask_last_m_chunk == 0 ) ? i_micro_kernel_config->datatype_size_in * i_vlen * i_m_blocking : i_micro_kernel_config->datatype_size_in * ( (i_vlen * (i_m_blocking-1)) + i_mask_last_m_chunk );
   LIBXSMM_UNUSED(i_mask_reg);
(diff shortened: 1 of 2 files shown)


src/generator_mateltwise_unary_binary_aarch64.c
```diff
@@ -550,6 +552,14 @@ void libxsmm_store_aarch64_2d_reg_block( libxsmm_generated_code*
     }
   }  else {
 #endif
+
+  if( l_is_sve ){
+    /* define predicate registers 0 and 1 for storing */
+    /* TODO: implement predication without re-initializing the predicate registers all the time */
+    libxsmm_generator_set_p_register_aarch64_sve( io_generated_code, 0, -1, i_gp_reg_mapping->gp_reg_scratch_0 );
+    libxsmm_generator_set_p_register_aarch64_sve( io_generated_code, 1, i_mask_last_m_chunk * i_micro_kernel_config->datatype_size_in, i_gp_reg_mapping->gp_reg_scratch_0 );
```
alheinecke, line 560: shouldn't that be datatype_size_out?

alheinecke:
actually the p registers should be set here:
https://github.com/libxsmm/libxsmm/blob/master/src/generator_mateltwise_unary_binary_aarch64.c#L2899
libxsmm_configure_unary_aarch64_kernel_vregs_masks

src/generator_mateltwise_unary_binary_aarch64.c
breuera: Yah, makes sense.
Guess there are a few other "datatype_size_in" which don't belong there as well; should fix them as well.

breuera:
Sounds good.
I am not a fan of the current solutions anyways..

breuera:
To avoid merge conflicts, the quick solution (with fixed output size) is now in #679 with arbitrary SVE sizes.

PR #679, opened 2022-07-30
Enabled SVE256 mateltwise (GeLU still requires a port).

breuera:


alheinecke (commented):
I will merge, I guess reduce transform, and transform TPPs and equations are also still missing.

PR #682, opened 2022-08-03
Fix amx gemm scratch size alloc

egeor:


src/generator_gemm_sse_avx_avx2_avx512.c
```diff
@@ -108,8 +108,10 @@ LIBXSMM_API_INTERN void libxsmm_generator_gemm_sse_avx_avx2_avx512_kernel_wrappe
 LIBXSMM_API_INTERN void libxsmm_generator_gemm_sse_avx_avx2_avx512_kernel( libxsmm_generated_code*        io_generated_code,
                                                                            libxsmm_loop_label_tracker*    io_loop_label_tracker,
                                                                            const libxsmm_gp_reg_mapping*  i_gp_reg_mapping,
-                                                                           const libxsmm_gemm_descriptor* i_xgemm_desc ) {
+                                                                           const libxsmm_gemm_descriptor* i_xgemm_desc_const ) {
   libxsmm_micro_kernel_config l_micro_kernel_config;
+  libxsmm_gemm_descriptor xgemm_desc_mod = *i_xgemm_desc_const;
+  libxsmm_gemm_descriptor *i_xgemm_desc = (libxsmm_gemm_descriptor*) &xgemm_desc_mod;
```
alheinecke, line 114: nit: local variables should start with l_

src/generator_gemm_sse_avx_avx2_avx512.c
```diff
@@ -108,8 +108,10 @@ LIBXSMM_API_INTERN void libxsmm_generator_gemm_sse_avx_avx2_avx512_kernel_wrappe
 LIBXSMM_API_INTERN void libxsmm_generator_gemm_sse_avx_avx2_avx512_kernel( libxsmm_generated_code*        io_generated_code,
                                                                            libxsmm_loop_label_tracker*    io_loop_label_tracker,
                                                                            const libxsmm_gp_reg_mapping*  i_gp_reg_mapping,
-                                                                           const libxsmm_gemm_descriptor* i_xgemm_desc ) {
+                                                                           const libxsmm_gemm_descriptor* i_xgemm_desc_const ) {
```
alheinecke, line 111: why renaming?

alheinecke (changes requested):
only a nit

src/generator_gemm_sse_avx_avx2_avx512.c
egeor: For the purposes of this (to restrict the scratch size when doing c vnni for non-amx and have it independent of ldc) I had to change the descriptor to use everywhere ldc = m for the gemm writing to scratch, and using the actial ldc only in the vnni cvt tpp at the end when storing from the scratch to the C buffer. Thus I had to create a new modifiable local xgemm descriptor (same thing that was done above for the transa  purposes to modify the descriptor)

alheinecke: so local,non-const variable with l_ and and keep the signature as it should fix it.

egeor: Then we'd have to change every single occurrence of the i_xgemm name that appears throughout this function to l_xgemm ... I am not saying this is not doable, but with the current code we change the name in the signature and nothing else.

If naming convention is more important then I can change it.

alheinecke: yes let's do this as I'm cleaning-up in every routine I see it. Just in order to improve quality, in vim "v" "%s/x/y" should do the job in a routine.

alheinecke (approved):

PR #687, opened 2022-08-11
BF8 (BR)GEMM on x86 via stack allocated tensors

egeor:
It enables BF8 (BR)GEMM on x86 (avx512 and amx) via stack allocated arrays.


samples/xgemm/kernel.c
```diff
@@ -63,9 +63,80 @@ typedef struct fusion_args {
   char *relu_bitmask;
 } fusion_args;
 
+
+#if 0
```
alheinecke, line 67: any reason to not clean-up?

samples/xgemm/kernel_test/bf8f32gemm_via_stack_nn_eqld.slurm
```diff
@@ -0,0 +1,65 @@
+#!/usr/bin/env bash
+export BF8_GEMM_VIA_STACK=1
```
alheinecke, line 2: Let's prepend LIBXSMM_

src/generator_common_x86.c
```diff
@@ -1958,6 +1958,17 @@ void libxsmm_generator_generic_loop_footer_with_idx_inc( libxsmm_generated_code*
   libxsmm_x86_instruction_jump_back_to_label( io_generated_code, LIBXSMM_X86_INSTR_JL, io_loop_label_tracker );
 }
 
+LIBXSMM_API_INTERN
+void libxsmm_generator_generic_loop_footer_with_idx_inc_reg_bound( libxsmm_generated_code*             io_generated_code,
```
alheinecke, line 1962: Can the simpler implementations call this one to reduce code duplication?

src/generator_gemm_common.c
```diff
@@ -1355,6 +1437,9 @@ void libxsmm_generator_gemm_init_micro_kernel_config_fullvector( libxsmm_micro_k
       io_micro_kernel_config->vector_length = 8;
       io_micro_kernel_config->datatype_size_in = 4;
       io_micro_kernel_config->datatype_size_out = 4;
+      if ( LIBXSMM_DATATYPE_BF8 == LIBXSMM_GETENUM_OUT( i_xgemm_desc->datatype ) ) {
```
alheinecke, line 1440: This seems to be dirty hack, can we catch in the code part which actually uses datatype_size_out and locally override?

src/generator_gemm_common.c
```diff
@@ -1584,6 +1669,9 @@ void libxsmm_generator_gemm_init_micro_kernel_config_fullvector( libxsmm_micro_k
       io_micro_kernel_config->vector_length = 16;
       io_micro_kernel_config->datatype_size_in = 4;
       io_micro_kernel_config->datatype_size_out = 4;
+      if ( LIBXSMM_DATATYPE_BF8 == LIBXSMM_GETENUM_OUT( i_xgemm_desc->datatype ) ) {
```
alheinecke, line 1672: dito

src/generator_gemm_sse_avx_avx2_avx512.c
```diff
@@ -105,13 +107,212 @@ LIBXSMM_API_INTERN void libxsmm_generator_gemm_sse_avx_avx2_avx512_kernel_wrappe
...
+}
+
 LIBXSMM_API_INTERN void libxsmm_generator_gemm_sse_avx_avx2_avx512_kernel( libxsmm_generated_code*        io_generated_code,
                                                                            libxsmm_loop_label_tracker*    io_loop_label_tracker,
                                                                            const libxsmm_gp_reg_mapping*  i_gp_reg_mapping,
                                                                            const libxsmm_gemm_descriptor* i_xgemm_desc ) {
   libxsmm_micro_kernel_config l_micro_kernel_config;
   libxsmm_gemm_descriptor l_xgemm_desc_mod = *i_xgemm_desc;
   libxsmm_gemm_descriptor *l_xgemm_desc = (libxsmm_gemm_descriptor*) &l_xgemm_desc_mod;
+  const char *const env_bf8_gemm_via_stack_alloc_tensors = getenv("BF8_GEMM_VIA_STACK");
```
alheinecke, line 314: Let's prepend LIBXSMM_

src/generator_gemm_sse_avx_avx2_avx512.c
```diff
@@ -1072,7 +1308,7 @@ LIBXSMM_API_INTERN unsigned int libxsmm_generator_gemm_sse_avx_avx2_avx512_get_m
       return 28;
     }
     /* handle bfoat8 on all AVX512 */
-    if ( ( LIBXSMM_DATATYPE_BF8 == LIBXSMM_GETENUM_INP( i_xgemm_desc->datatype ) ) ) {
+    if ( ( LIBXSMM_DATATYPE_BF8 == LIBXSMM_GETENUM_INP( i_xgemm_desc->datatype ) ) || ( LIBXSMM_DATATYPE_BF8 == LIBXSMM_GETENUM_OUT( i_xgemm_desc->datatype ) ) ) {
```
alheinecke, line 1311: Why? There is no emulation need, so all 30 registers should be available.

alheinecke (commented):
Minor nits.

Let's see how we can get AVX512 / AVX512_VL256 vnni4->vnni2 and vnni4->norm routines ASAP, then we can merge.

samples/xgemm/kernel.c
egeor: In case we don't want to lose the scalar sigmoid implementation... We can take this out otherwise

src/generator_common_x86.c
egeor: The simpler ones don't require the extra reg for the cmp... we could do the same, but then we would have to provide in the args yet another reg

src/generator_gemm_common.c
egeor: This is used throughout in the load/store C gemm functions, not sure if it is better to override there?

src/generator_gemm_sse_avx_avx2_avx512.c
egeor: I believe otherwise the bf8 cvt is broken here (it hardcodes vreg 2 and 3 for tmp): https://github.com/libxsmm/libxsmm/blob/main/src/generator_gemm_common.c#L2991

alheinecke: and I guess as we are now using FP32 GEMM, there are cases where we go into 1D blocking which fully unrolls in N?

src/generator_gemm_common.c
alheinecke: Hmm, so this is because of adding internally a F32BF8 kernel such that we don't need to allocate C on the stack and can directly store BF8, correct? Can we add a comment?
My fear is the entanglement this side-band causes and that we oversee this when rewriting the Spaghetti code of libxsmm_generator_gemm_init_micro_kernel_config_fullvector. 
Orthogonal, should we enabled just for fun GEMMs like F32BF8, F32BF16 etc. (not in this PR)?!

src/generator_gemm_sse_avx_avx2_avx512.c
egeor: I believe so

src/generator_common_x86.c
alheinecke: I see, then let's fix this when unifying eltwise and GEMM aux routines.

samples/xgemm/kernel.c
alheinecke: Let's save the scalar implementation externally and remove.

src/generator_gemm_common.c
egeor: Exactly!
The right thing (making it look less hacky?) would be basically to make another outer if before this point checking in ==  f32 and out == bf8 (instead of adding on the spot a change, because now the underlying assumption is when in == f32 then always out == f32). Of course for now it would go only on this esoteric case when emulating bf8, but as you say it would allow to enable (for fun hehe) weird GEMMs like F32BF8 etc

alheinecke (approved):

PR #691, opened 2022-08-26
bf8 support added to FC layer test code

nkmellem:


alheinecke:
thanks. Should we wait for the convolutions as well before reviewing & merging?

hfp:
This PR is transferred to https://github.com/libxsmm/libxsmm_dnn/pull/1 with some minor touch-up. I also preserved the original committer/contribution.

PR #701, opened 2022-10-19
Avx2 bf16 tpp support

egeor:


src/generator_mateltwise_reduce_avx_avx512.c
```diff
@@ -23,6 +23,90 @@
 # define LIBXSMM_GENERATOR_MATELTWISE_REDUCE_AVX_AVX512_JUMP_LABEL_TRACKER_MALLOC
 #endif
 
+LIBXSMM_API_INTERN
```
alheinecke, line 26: nice, some clean-up :-)

samples/eltwise/eltwise_unary_simple.c
```diff
@@ -413,7 +413,7 @@ int test_unary_op( const libxsmm_blasint M, const libxsmm_blasint N, const libxs
   printf("Check-norm    : %.24f\n\n", norms_out.normf_rel);
 
   if ( op == RCP_OP || op == RCP_SQRT_OP ) {
-    error_bound = 0.0027;
+    error_bound = 0.008;
```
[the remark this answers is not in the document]
egeor: OK can do

samples/eltwise/eltwise_unary_reduce.c
```diff
@@ -264,6 +264,9 @@ void setup_tpp_kernel_and_param_struct( libxsmm_meltwfunction_unary *res_kernel,
     } else {
       unary_flags = LIBXSMM_MELTW_FLAG_UNARY_IDX_SIZE_4BYTES;
     }
+    if (reduce_on_outputs > 0) {
```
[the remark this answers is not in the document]
egeor: It's not in master (we use the default setup for Embedding and as such we unit test only reduce_on_outputs == 0), but since we support this functionality in the JITer I added it in the driver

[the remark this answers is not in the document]
alheinecke: Ack

alheinecke:
Is it ready for final review and merge?

egeor:
Yep, I think it can be merged

samples/eltwise/eltwise_unary_simple.c
```diff
@@ -413,7 +413,11 @@ int test_unary_op( const libxsmm_blasint M, const libxsmm_blasint N, const libxs
   printf("Check-norm    : %.24f\n\n", norms_out.normf_rel);
 
   if ( op == RCP_OP || op == RCP_SQRT_OP ) {
-    error_bound = 0.0027;
+    if (dtype_in != LIBXSMM_DATATYPE_F32 || dtype_out != LIBXSMM_DATATYPE_F32) {
```
alheinecke, line 416: Should that be bound to F16? (BF16 was also fine with the higher bar)

alheinecke (commented):
Minor Nit with the correctness check

samples/eltwise/eltwise_unary_simple.c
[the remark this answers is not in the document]
alheinecke: some update below

samples/eltwise/eltwise_unary_simple.c
egeor: Well, the issue is that the avx2 codepath is using different approx instructions for these ops. This approx discrepancy seems to manifest as higher error bars now that we added bf16/f16 for avx2

alheinecke: can we add some guarding similar to what we do in dropout:
https://github.com/libxsmm/libxsmm/blob/main/samples/eltwise/eltwise_unary_dropout.c#L73
Then we can keep the error bars per arch as low as possible.

egeor: I guess what you really mean is to "enhance" the newly introduced continue with && libxsmm_get_target_archid() <= avx2 or so, right?

alheinecke: yepp.. so that the old value is used for everything else than AVX2 FP16 and the new one is only used for AVX2 FP16/BF16 (not sure if both FP16/BF16 creating the issue)

egeor: Don't quite recall if it is only BF16 or both. For now I change the bound only for bf16 cases and see what the CI says...

alheinecke (approved):

PR #703, opened 2022-10-30
x86 fp64 basic tpp support

egeor:


src/generator_mateltwise_reduce_avx_avx512.c
```diff
@@ -737,6 +747,10 @@ void libxsmm_generator_reduce_cols_avx512_microkernel( libxsmm_generated_code*
     }
   }
 
+  if (LIBXSMM_DATATYPE_F64 == libxsmm_meltw_getenum_precision(i_mateltwise_desc, LIBXSMM_MELTW_FIELD_COMP)) {
```
alheinecke, line 750: Should that be handled when setting up the vlen in first place?

src/generator_mateltwise_unary_binary_avx_avx512.c
```diff
@@ -2965,6 +3015,11 @@ void libxsmm_generator_unary_binary_avx512_microkernel( libxsmm_generated_code*
   }
 
   /* Configure vlens */
+  if (LIBXSMM_DATATYPE_F64 == libxsmm_meltw_getenum_precision(i_mateltwise_desc, LIBXSMM_MELTW_FIELD_COMP)) {
+    i_micro_kernel_config->use_fp64_compute = 1;
```
alheinecke, line 3019: Why having this flag, when most the code places do the right thing check for compute datatype being FP64? This is just redundant information so we should not do this.

src/generator_matequation_regblocks_avx_avx512.c
```diff
@@ -772,32 +792,45 @@ void libxsmm_generator_mateqn_compute_unary_op_2d_reg_block( libxsmm_generated_c
     for (im = 0; im < i_m_blocking; im++) {
       cur_vreg = i_start_vreg + in * i_m_blocking + im;
       if (i_op_type == LIBXSMM_MELTW_TYPE_UNARY_X2) {
-        libxsmm_x86_instruction_vec_compute_3reg( io_generated_code, LIBXSMM_X86_INSTR_VMULPS, i_micro_kernel_config->vector_name, cur_vreg, cur_vreg, cur_vreg );
+        libxsmm_x86_instruction_vec_compute_3reg( io_generated_code, (i_meqn_micro_kernel_config->use_fp64_compute > 0) ? LIBXSMM_X86_INSTR_VMULPD : LIBXSMM_X86_INSTR_VMULPS, i_micro_kernel_config->vector_name, cur_vreg, cur_vreg, cur_vreg );
```
alheinecke, line 795: see comment above... we need to clean-up the code, so create now two ways of checking is just creating more work down the road.

alheinecke (commented):
2 things:
- the Fp64 compute precision should be handled identically in all case: query the datatype field
- We still don't cover all possible input arch. It seems CPX/SPR will just run F32 unit tests. Hence I suggested the sourcing previously. Then we have the precision code in on place and a one liner in each test file. That should reduce duplication and therefore room for error.

src/generator_mateltwise_reduce_avx_avx512.c
egeor: Yep, it can be moved over there. Regardless though, as we discussed this part of the code needs to be cleaned up/re-written such that vlen is setup by a routine rather than a sequence of dispersed if conditions.

src/generator_mateltwise_unary_binary_avx_avx512.c
egeor: The reason is that (as we have discussed in the past) some routines have incomplete/broken API, even more as we add more and more functionalities/support. For example, the "libxsmm_configure_unary_kernel_vregs_masks" used later needs to know the compute precision to setup properly some vregs, but the descriptor is not part of its API (whereas the kernel_config is). IMHO, we should revisit *all* routines and redesign/clean their API to be future proof as we discussed in the past, but this is a big/different overhaul which is not part of this PR. Changing this particular API is also doable but does not fix the issue in the grand scheme of things.

src/generator_matequation_regblocks_avx_avx512.c
egeor: same issue as explained in the comments above...

egeor:
> 2 things:
> 
> * the Fp64 compute precision should be handled identically in all case: query the datatype field

See my separate comments. In principle some routines should change API to get the descriptor as an input. Should be easily doable for this specific use-case.

> * We still don't cover all possible input arch. It seems CPX/SPR will just run F32 unit tests. Hence I suggested the sourcing previously. Then we have the precision code in on place and a one liner in each test file. That should reduce duplication and therefore room for error.

src/generator_matequation_regblocks_avx_avx512.c
alheinecke: yepp. see bigger comment below.

src/generator_mateltwise_unary_binary_avx_avx512.c
alheinecke: yepp, this is needed. I feel it's time to do the clean-up now before creating more code which is not streamlined.

alheinecke:
I think we need a plan for fixing the infrastructure. Right now we are just piling on more hacks. I think it's fine to merge this, as it's done. But this should be the last PR which adds new arch/precision; we first need to fix the infrastructure.

alheinecke (commented):
Let's move:
```
if
 [[ ${LIBXSMM_TARGET} == "snb" ]]; then
    exit 0;
fi
if [[ ${LIBXSMM_TARGET} == "wsm" ]]; then
    exit 0;
fi
```

also in the script. Or is there a reason, why it stayed in in each of the testers... (in case see that exit 0 in source doesn't kill the script..., but there should be a way to achieve this.... @hfp: any ideas if this is the issue?)

egeor:
These ifs per se do not setup any precision, they just exit so not sure why they belong in the setup prec script in the first place. If the exit works in that script I am also fine moving those there

alheinecke:
Yeah, let's try to move them there as with SSE support in future they will move over there anyways.. Perhaps just setting an "empty" precision string is fine as well.

hfp:
> Let's move:
> 
> ```
> if
>  [[ ${LIBXSMM_TARGET} == "snb" ]]; then
>     exit 0;
> fi
> if [[ ${LIBXSMM_TARGET} == "wsm" ]]; then
>     exit 0;
> fi
> ```
> 
> also in the script. Or is there a reason, why it stayed in in each of the testers... (in case see that exit 0 in source doesn't kill the script..., but there should be a way to achieve this.... @hfp: any ideas if this is the issue?)

I will take a look, i.e., if `exit` is an issue when executed for instance per `tool_pexec` etc.

alheinecke (commented):
Why is setup_tpp_prec_list 2x in this PR. One time in the samples/eltwise folder and the other on in samples/eltwise/kernel_test. I think the one in kernel_test should be enough, shouldn't it?

egeor:
Since the run scripts are executed from : samples/eltwise I think the one there should be kept

alheinecke (approved):

PR #706, opened 2022-11-25
BFDOT aarch64-SVE-NEOV1 support

alheinecke:
This PR adds support for the BFDOT instruction for performance comparison compared to BFMMLA on the Neoverse V1 CPU. Additionally, the cpuid APIs have been extended to provide querying of the VNNI/BFDOT/BFMMLA packing factor.


src/libxsmm_cpuid_x86.c
```diff
@@ -426,3 +427,48 @@ LIBXSMM_API int libxsmm_cpuid_vlen32(int id)
   }
   return result;
 }
+
+LIBXSMM_API int libxsmm_cpuid_dot_pack_factor(libxsmm_datatype in_dtype)
+{
+  int result = -1;
```
ddkalamk, line 433: Why default is -1 instead of 1?

alheinecke:
for invalid as we don't need packing, but I guess your point is that one then needs an if in user code on the datatype, so with 1 this if on user side would not be needed, correct?

src/libxsmm_cpuid_x86.c
alheinecke: for invalid as we don't need packing, but I guess your point is that one then needs an if in user code on the datatype, so with 1 this if on user side would not be needed, correct?

ddkalamk: We may still need if condition to decide whether to pack or not but intuitively I would return pack size of 1 instead of -1.

alheinecke: hmm... yeah, 1 means no packing and -1 means no packing as well. but 1 allows to be easily used in computations when the if can be avoided. Let me change this with next commit.

src/generator_gemm.c
```diff
@@ -211,7 +212,39 @@ void libxsmm_generator_gemm_kernel( libxsmm_generated_code*        io_generated_
   } else if ( ( io_generated_code->arch >= LIBXSMM_AARCH64_V81    )  &&
               ( io_generated_code->arch <= LIBXSMM_AARCH64_ALLFEAT ) &&
               (    ( LIBXSMM_DATATYPE_BF16 == LIBXSMM_GETENUM_INP( l_xgemm_desc_mod.datatype ) )
-                || ( LIBXSMM_DATATYPE_I8   == LIBXSMM_GETENUM_INP( l_xgemm_desc_mod.datatype ) ) ) ) {
+                || ( LIBXSMM_DATATYPE_I8   == LIBXSMM_GETENUM_INP( l_xgemm_desc_mod.datatype ) ) ) &&
+              ( l_aarch64_bfdot != 0 ) ) {
+    /* TODO (MMLA): add flags and check on MMLA-formated A/B */
```
egeor, line 217: For clarity we may want to remove these comments since this is not the MMLA branch of the if condition

src/generator_gemm_aarch64.c
```diff
@@ -1103,18 +1135,28 @@ void libxsmm_generator_gemm_aarch64_kernel( libxsmm_generated_code*        io_ge
...
   /* enable MMLA settings for supported datatypes */
-  if (    LIBXSMM_DATATYPE_BF16 == LIBXSMM_GETENUM_INP( i_xgemm_desc->datatype )
-      || LIBXSMM_DATATYPE_I8   == LIBXSMM_GETENUM_INP( i_xgemm_desc->datatype ) ) {
-    l_use_mmla = 1;
+  if ( (LIBXSMM_DATATYPE_BF16 == LIBXSMM_GETENUM_INP( i_xgemm_desc->datatype )) ||
+       (LIBXSMM_DATATYPE_I8   == LIBXSMM_GETENUM_INP( i_xgemm_desc->datatype )) ) {
+    if ( l_use_bfdot == 0 ) {
+      l_use_mmla = 1;
+    } else {
+      if ( l_use_bfdot != 0 ) {
```
egeor, line 1148: Shouldn't this be just l_use_mmla = 0 ?

egeor:
Other than the two minor nits I commented it looks good to me!

src/generator_gemm.c
alheinecke: yes the comment needs to BFDOT

PR #707, opened 2022-11-30
Add dump unary tpp op

egeor:
This PR introduces a DUMP operator which allows to dump intermediate results in TPP Equations


samples/equation/equation_gather_reduce.c
```diff
@@ -31,7 +31,7 @@ void eqn1_f32f32(libxsmm_blasint M, libxsmm_blasint N, libxsmm_blasint ld, float
 
 int main( int argc, char* argv[] ) {
   int ret = EXIT_SUCCESS;
-  double error_bound = 0.00001;
+  double error_bound = 0.0001;
```
alheinecke, line 34: Why do we need to reduce this when just dumping a node?

alheinecke (commented):
minor question on unit test

samples/equation/equation_gather_reduce.c
egeor: Oh this has nothing to do with this PR, this "numerical issue" was just revealed in this CI run... I.e. the same problem is also in the master (and will be fixed when merged). The issue btw is the same as the layernorm thingy... compiling with -O0 the reference gives bitwise identical results as the TPP code. Without optimization level the error bound varies (this equation essentially performs a reduce cols idx and different accumulation orders may be triggered based on different compiler opt levels...)

PR #708, opened 2022-12-02
16-bit transforms norm_to_vnni4 and vnni4_to_vnni4T for avx512 and avx512_vl256

egeor:
This PR introduces the 16-bit transforms norm_to_vnni4 and vnni4_to_vnni4T for avx512 and avx512_vl256


alheinecke (approved):
le

egeor:
The CI went through for the scalar codes on AVX/SSE so we should be good to merge

alheinecke (commented):
Do you still want put the norm_to_vnni4t TPP and then calling 64bit normal transpose inside before merging?

alheinecke:
@hfp: please advise when main is functional again, so that we can merge

egeor:
> Do you still want put the norm_to_vnni4t TPP and then calling 64bit normal transpose inside before merging?

OK will add it

egeor:
Now the norm_to_vnni4T 16bit transpose is also in and we can merge back

alheinecke (approved):

PR #709, opened 2022-12-03
Align x86 VNNI implementation with aarch64's BFDOT/BFMMLA 

alheinecke:


alheinecke (commented):
let's wait for SA, some minor touch ups will be needed.

PR #712, opened 2022-12-10
refactors for the x86 SIMD kloop kernel selection

alheinecke:


src/generator_gemm_avx512_microkernel.c
```diff
@@ -12,6 +12,65 @@
...
+      for ( l_k = 0; l_k < i_k_blocking; l_k += l_k_pack_factor) {
+        l_generator_microkernel(io_generated_code, i_gp_reg_mapping, i_micro_kernel_config,
+                                i_xgemm_desc, i_m_blocking, i_n_blocking, (int)l_k);
+      }
+    } else {
+      for ( l_k = 0; l_k < i_k_blocking; l_k += l_k_pack_factor) {
+        l_generator_microkernel(io_generated_code, i_gp_reg_mapping, i_micro_kernel_config,
+                                i_xgemm_desc, i_m_blocking, i_n_blocking, -1);
+      }
+    }
```
egeor, line 70: nit: there is code duplication here, whereas merely the code is calling the same function with same args except the last one depending on the condition "i_k_blocking == (unsigned int)i_xgemm_desc->k"

src/generator_gemm_avx_microkernel.c
```diff
@@ -14,13 +14,42 @@
...
+
+  if ( i_k_blocking == (unsigned int)i_xgemm_desc->k ) {
+    for ( l_k = 0; l_k < i_k_blocking; l_k += l_k_pack_factor) {
+      libxsmm_generator_gemm_avx_microkernel(io_generated_code, i_gp_reg_mapping, i_micro_kernel_config,
+                                             i_xgemm_desc, i_m_blocking, i_n_blocking, (int)l_k);
+    }
+  } else {
+    for ( l_k = 0; l_k < i_k_blocking; l_k += l_k_pack_factor) {
+      libxsmm_generator_gemm_avx_microkernel(io_generated_code, i_gp_reg_mapping, i_micro_kernel_config,
+                                             i_xgemm_desc, i_m_blocking, i_n_blocking, -1);
```
egeor, line 40: nit: there is code duplication here, whereas merely the code is calling the same function with same args except the last one depending on the condition "i_k_blocking == (unsigned int)i_xgemm_desc->k"

src/generator_gemm_avx2_microkernel.c
```diff
@@ -264,126 +145,218 @@ void libxsmm_generator_gemm_avx2_microkernel_bf16_vnni_emu( libxsmm_generated_co
...
+
+        libxsmm_x86_instruction_vec_move( io_generated_code,
+                                      i_micro_kernel_config->instruction_set,
+                                      i_micro_kernel_config->b_vmove_instruction,
+                                      i_gp_reg_mapping->gp_reg_b,
+                                      LIBXSMM_X86_GP_REG_UNDEF, 0,
+                                      l_b_offset,
+                                      i_micro_kernel_config->vector_name,
+                                      l_n, 0, 1, 0 );
+      }
```
egeor, line 227: Nit: I know this issue was not introduced per se with this refactoring, but there is also quite some code duplication here depending on the i_offset value... the l_n loops are literately identical, only the l_b_offsets vary in the first term (being zero in the k non-unrolled case). Of course the gp_reg_b advancement would have to happen only in the non-unrolled version.

egeor:
Looks good to me, just a couple of nits pertaining to superfluous code duplication.

src/generator_gemm_avx_microkernel.c
alheinecke: fixed

src/generator_gemm_avx512_microkernel.c
alheinecke: fixed

src/generator_gemm_avx2_microkernel.c
```diff
@@ -14,172 +14,58 @@
...
-      libxsmm_x86_instruction_vec_compute_3reg( io_generated_code,
-                                                LIBXSMM_X86_INSTR_VPADDD,
-                                                i_micro_kernel_config->vector_name,
-                                                i_n_blocking,
-                                                l_vec_reg_acc_start + l_m + (l_m_blocking * l_n),
-                                                l_vec_reg_acc_start + l_m + (l_m_blocking * l_n) );
+    for ( l_k = 0; l_k < i_k_blocking; l_k += l_k_pack_factor) {
+      l_generator_microkernel(io_generated_code, i_gp_reg_mapping, i_micro_kernel_config,
+                              i_xgemm_desc, i_m_blocking, i_n_blocking, -1);
     }
```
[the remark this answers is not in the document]
alheinecke: fixed

PR #713, opened 2022-12-11
Fix GEMM and eltwise/equation Tests

alheinecke:
Recently we introduced ignore list for failing tests, however they are very aggressive (switching to easily "green"). Also, some scripts just return "success". This PR tracks the important and urgent task carried out in https://github.com/libxsmm/libxsmm/tree/fix_tests to establish rigorous unit testing and skipping as little tests as possible.


hfp:
Equation tests at least suffer from numerical instabilities perhaps related to compiler generated gold code, i.e., randomized compiler, optimizations/vectorization/scalar code. For instance, `./equation_layernorm 2 99 1 0 3 0` gave an error margin of `0.001102204469695817036848` (`avx512_vl256_clx`). We have now separate margins for F32, Bf16, S1=S3=1,  S2=1, however, the mentioned/recent case is yet again different (S3=1, 64<S2). I will rework again but ultimately @egeor should judge the cases.

hfp (approved):
Very nice, I just did not realize how much of the following we have:

```bash
if [[ ${LIBXSMM_TARGET} == "snb" ]]; then
  exit 0;
fi
```

( Further, we never covered alternative code-path names aka "sse42" vs "wsm". )

I will pull this PR locally and play with it a bit more and add/update whitelists.

hfp:
KNM (as opposed to KNL) has whitelisted even case like `dgemm_nn_eqld`, `dgemm_nn_gtld`, `dgemm_nt_eqld`, `dgemm_nt_gtld`, and `dgemm_tn_eqld`. We have cases with failing to JIT as kernel and errors like `LIBXSMM ERROR (libxsmm_generator_gemm_avx512_microkernel_fsdbcst_qfma): invalid N blocking in microkernel (error #90016)!`.

egeor:
>  We have now separate margins for F32, Bf16, S1=S3=1, S2=1, however, the mentioned/recent case is yet again different (S3=1, 64<S2). I will rework again but ultimately @egeor should judge the cases.

@hfp I believe our best-effort thing here would be to run exhaustively 100x100x100 (S1xS2xS3) experiments, potentially with icc, gcc and clang, and from there extract weaker norm bounds and potentially loose ones only under certain scenarios as the one you mentioned above. Could you take a stab at this?

hfp:
> please rework SKIPPED,whitelisting, it's not correct that crashes, JIT failures etc. are turned to "green" only thing allowed to be green is JIT pointer == NULL

It was by design to keep things green while working on this; not sure when to tighten the screws. Indeed, I need to introduce perhaps a 2nd column capturing the return code and then have some flexibility in the runner to mark, e.g., `132` (aka crash) as failure. However, we do not have strong grip on values of return codes, i.e., `exit(-1)` (and friends) are scattered over the whole code base.

Ideally, LIBXSMM's code would never call exit directly (perhaps via assert). Even without 2nd column, we can build/run/test everything twice one time built with `SYM=2` and one time built without it (this would yield the new "BLACK" state for whitelisted case that suddenly would not fail anymore).

hfp:
The backend now consistently uses `LIBXSMM_EXIT_ERROR()` instead of `exit(-1)`. This new convention is enforced by `scripts/tool_normalize.sh`. Most of our drivers continue using `exit(-1)`, which in fact is used to terminate if JIT-codegen returns a NULL-pointer ("JIT bailing").

The `scripts/tool_pexec.sh` now supports `--xfail` (`-x`) command line flag, or `PEXEC_XF` environment variable, which allows to consider exit codes `2..254` to be considered failures (even if the case is whitelisted).

We can deploy a different default for "XFAIL" at a later point in time, or when we fixed most expected issues upfront. We can also deploy it once this PR is merged.

alheinecke:
Can you please prioritize adding a new return code for JIT-Failure in this branch/PR, let's say "42". We are only allowed to declare a "42" passing, when it's in the SKIP list. If it's in the SKIP list and it crashes, gives wrong numerical result etc. it MUST fail.

This is the reasons, why we have a hard time debugging right now what is going on. All cases incl. bugs are set to "PASS", that's why you last commit was not "helpful". The equations for AVX2 were enabled back, as we believe they should all work, except split SGD. So we cannot simply put failing tests into SKIP list. Every single cases needs to be understood separately.

hfp:
> Can you please prioritize adding a new return code for JIT-Failure in this branch/PR, let's say "42". We are only allowed to declare a "42" passing, when it's in the SKIP list. If it's in the SKIP list and it crashes, gives wrong numerical result etc. it MUST fail.
> 
> This is the reasons, why we have a hard time debugging right now what is going on. All cases incl. bugs are set to "PASS", that's why you last commit was not "helpful". The equations for AVX2 were enabled back, as we believe they should all work, except split SGD. So we cannot simply put failing tests into SKIP list. Every single cases needs to be understood separately.

Yes, that's we have here; return code is `15` or `0xF`. Code generator exits with this code instead of `-1`. I can enable "xfail" for this branch (`tool_pexec.sh`), and we can then look at every single case of "bailing out" in the code generator. Whitelists do not matter for xfail of exit code is 2..254. Of course, we can cleanup whitelists later when we scrubbed our backend.

hfp:
If it now passes "green", I will toggle what's accepted (regardless of what is listed "white") and we should see the true status, i.e., assertions in the code generator ("backend").

alheinecke:
I just review the code, this commit https://github.com/libxsmm/libxsmm/commit/4c7d8cf5ea1a9f0606ebc2d371e82e063266bd20 is still not doing the right. When this asserting is triggered, it should turn red, correct?
The only case allowed for green is 
- entry ALLOW/SKIP list
- NULL pointer returned from user visible JITer, e.g. https://github.com/libxsmm/libxsmm/blob/main/samples/xgemm/kernel.c#L1470

All other case need returning non-zero exit code most turn the CI red.

hfp:
> I just review the code, this commit [4c7d8cf](https://github.com/libxsmm/libxsmm/commit/4c7d8cf5ea1a9f0606ebc2d371e82e063266bd20) is still not doing the right thing, I'm afraid. When this asserting is triggered, it should turn red, correct?

Yes, if the pexec tool considers 2..254 as failure. Next commit will do this.

> The only case allowed for green is
> 
>     * entry in ALLOW/SKIP list

ACK.

>     * NULL pointer returned from user visible JITer, e.g. https://github.com/libxsmm/libxsmm/blob/main/samples/xgemm/kernel.c#L1470

Well, that's xgemm but in general we have no way to detect "NULL pointer returned" from a CI-perspective.

> All other cases returning non-zero exit code MUST turn the CI red.

ACK. Same condition like XFAIL.

hfp:
3h gap on my end before continuing. Current status: KNM is instable, and AVX512-VL256 needs updated whitelists, then we can turn on "true status".

alheinecke:
@hfp: why are intrinsic fixes needed? The code should be free of any intrinsics.

hfp:
> @hfp: why are intrinsic fixes needed? The code should be free of any intrinsics.

The test based on Intel Compiler uses "current" and "random" compiler, the latter potentially an older version than current. It turned out selecting an old/legacy version. Also, The `__AVX512F__`-protected section in the driver code used a pseudo intrinsic only available under a slightly different condition (`LIBXSMM_INTRINSICS_AVX512` is defined only if the compiler is able to generate this code without special flags).

alheinecke:
OK... so we also need to clean-up a driver

alheinecke:
Ah... it's the equation stuff... I would not fix anything on the AVX512 side in the driver @egeor  I guess we can delete the intrinisc codes, correct? We just need to get the error bars for the scalar code correct....

egeor:
Well, the equation drivers have this intrinsics-vectorized code for performance comparisons on avx512 platforms (see SC21 TPP paper). These drivers guard the intrinsics path under some #ifdef and use scalar code if that is not defined (in order e.g. to check correctness on arm, avx2 etc)

alheinecke:
@egeor : since we have the perf test in the SC21 branch and they are only AVX512, what's the value of keeping them?

egeor:
We can remove, but in that sense we can remove tons of other code too that lies in prior papers/branches :)

alheinecke:
YES. Let's remove code which is not part of the main library and not needed. What do you have in mind?

alheinecke:
Why can't we do regression testing against our own code, when we log perf now and then remove. For ASIMD, AVX2, equations or eltwise and GEMM in general we don't have this and never planned to do, so why is this important for equations AVX512 then?

egeor:
The point of this specific was not so much about regression but rather comparing the eqs machinery with something decent (albeit only avx512). Agreed we don't have equivalent for other archs TPPs. I guess we can remove then.

alheinecke:
@egeor : ` LIBXSMM_TARGET=snb ./kernel_test/sgemm_nn_eqld_bop0_uop3_cvnni0.slurm` is failing with your patches, but it was passing with: 6928f77bd8bcffa8f263036e91713d85a6fd6483 
It seems the condition you added are too strict, can you revise?

src/generator_common.c
```diff
@@ -1293,6 +1290,66 @@ LIBXSMM_API_INTERN unsigned int libxsmm_compute_equalized_blocking(
   return l_ret;
 }
 
+LIBXSMM_API_INTERN
```
alheinecke, line 1293: shouldn't we move these functions into a matelwise_common file

src/generator_common.h
```diff
@@ -2125,6 +2128,12 @@ typedef enum libxsmm_ulp_precision {
   LIBXSMM_ULP_PRECISION_ESTIMATE /* can be pretty bad, but should have the correct order of magnitude */
 } libxsmm_ulp_precision;
 
+LIBXSMM_API_INTERN
```
alheinecke, line 2131: should they go into a mateltwise common file?

src/generator_mateltwise_sse_avx_avx512.c
```diff
@@ -632,13 +632,83 @@ void libxsmm_generator_mateltwise_init_micro_kernel_config_fullvector( libxsmm_g
...
+                  (i_mateltwise_desc->param == LIBXSMM_MELTW_TYPE_UNARY_REDUCE_X_OP_ADD)   ||
+                  (i_mateltwise_desc->param == LIBXSMM_MELTW_TYPE_UNARY_REDUCE_X2_OP_ADD)    ||
+                  (i_mateltwise_desc->param == LIBXSMM_MELTW_TYPE_UNARY_REDUCE_X_X2_OP_ADD) ||
+                  (i_mateltwise_desc->param == LIBXSMM_MELTW_TYPE_UNARY_DUMP)         ||
+                  (i_mateltwise_desc->param == LIBXSMM_MELTW_TYPE_UNARY_REDUCE_X_OP_MAX))       ) ? 1 : 0;
+  unsigned int has_inp_or_out_fp8 = (libxsmm_generator_mateltwise_involves_prec(i_mateltwise_desc, LIBXSMM_DATATYPE_BF8) > 0 || libxsmm_generator_mateltwise_involves_prec(i_mateltwise_desc, LIBXSMM_DATATYPE_HF8) > 0) ? 1 : 0;
+  unsigned int has_inp_or_out_fp64 = (libxsmm_generator_mateltwise_involves_prec(i_mateltwise_desc, LIBXSMM_DATATYPE_F64) > 0) ? 1 : 0;
+  unsigned int has_all_inp_and_out_fp64 = (libxsmm_generator_mateltwise_all_inp_comp_out_prec(i_mateltwise_desc, LIBXSMM_DATATYPE_F64) > 0) ? 1 : 0;
+
+  if ((io_generated_code->arch < LIBXSMM_X86_AVX) && (is_transform_tpp == 0)) {
```
alheinecke, line 675: Isn't here now a condition missing that GELU on AVX is avoided?

alheinecke (commented):
some minor nits

samples/xgemm/kernel.c
```diff
@@ -133,6 +133,10 @@ float fsigmoid(float x) {
   libxsmm_meltwfunction_unary unary_kernel  = libxsmm_dispatch_meltw_unary_v2( LIBXSMM_MELTW_TYPE_UNARY_SIGMOID, unary_shape, LIBXSMM_MELTW_FLAG_UNARY_NONE );
   libxsmm_meltw_unary_param unary_param;
   float in = x, out;
+  if( unary_kernel == 0 ) {
+    printf("JIT failed, please run with LIBXSMM_VERBOSE=-1 and/or with debug mode LIBXSMM library!\n");
+    exit(EXIT_FAILURE);
```
alheinecke, line 138: @hfp we need to put a different exit code. That is also true the GEMM JIT.

hfp: I don't think so. This is "user code" finally discovering a NULL-pointer. I am assuming there is no previous assertion or exit from inside of our library code (libxsmm). This case shall be whitelisted aka allowed to fail. I mean, we now want to collect this case in a whitelist and to be recorded as a potential gap in our implementation (no code generated). It was a graceful error going all the way through our library's code and producing a NULL-pointer (yeahh!).

I think the current convention is:

* `EXIT_FAILURE`: missing error margin (validation failed); also for incorrect input (command line arguments, etc).
* `-1`: NULL-pointer discovered.

... both of which can be whitelisted.

The range in between the two, i.e., `2..254` is "reserved" for accidents happening at process level aka SEGFAULT, DIVBYZERO, SIGILL, etc. Such error cannot be whitelisted (at least not by our new default in `tool_pexec.sh`).

src/generator_common.c
egeor: Done

src/generator_common.h
egeor: Done

src/generator_mateltwise_sse_avx_avx512.c
egeor: Indeed oversight, done also for eqs

alheinecke:
@hfp : for now the GEMM test return 15 for JIT=NULL which is OK, but it's set to ERROR, how can they be set to ALLOW?

hfp:
> @hfp : for now the GEMM test return 15 for JIT=NULL which is OK, but it's set to ERROR, how can they be set to ALLOW? Or to ask differently, why are they return F and not -1?

We have to decide what/if we want to whitelist anything. You wanted to return code of say `42` in the driver if it discovered a NULL-pointer (I chose `15`). I can certainly revert to `-1`. Alternatively, we can allow whitelisting any return code including `2..254`.

What I will do meanwhile:  
I will re-enable the option to allow/whitelist all return codes including `2..254`. Originally, we wanted to allow gaps but thereby document them using an allow-list. We should keep the ability to distinct `-1`, `+1`, and `15`, i.e., I re-enable the ability to whitelist what we want as opposed to always failing for `2..254`. I must remove, update, or at least revisit all whitelists.

alheinecke:
please don't change the allow listing for GEMMs they are correct and everything works. Nothing needs to be modified. The numbers (the actual return code we use are irrelevant). The only thing what is broken right now is the allow listing mechanism. We have simply 4 cases:

- success -> return code "0"
- JIT fails -> return code "-1"  -> right now the GEMM test return here 15
- crash/segfault -> no return code
- wrong results -> return for now "1-255" or so, doesn't really matter.

Handling of the return value:
When non-zero: when -1/"JIT Fails" check if the case is in the allow list. If yes -> mark allow, if no -> raise error.
When zero: check for allow listing: if case is in there -> raise error as this cannot be. if no -> mark success.

alheinecke:
With this mechanism, the allow listing automatically become our tool of tracking implementation holes for arch + precision combination. This is what we want.

hfp:
> With this mechanism, the allow listing automatically become our tool of tracking implementation holes for arch + precision combination. This is what we want.

Exactly.

hfp:
> please don't change the allow listing for GEMMs they are correct and everything works. Nothing needs to be modified. The numbers the actual return code we use are irrelevant. The only thing what is broken right now is the allow listing mechanism. We have simply 4 cases:
> 
>     * success -> return code "0"
ACK.

>     * JIT fails -> return code "-1"  -> right now the GEMM test return here 15
You wanted something other than `-1` if I understood this correctly from previous comment. The thing is, `-1` is used a lot to exit in case of incorrect command line arguments, etc. Anyway, I can change it from `15` back to `-1` in case of NULL-pointer (driver codes).

>     * crash/segfault -> no return code
We always get a return code; that's an OS-mechanism. The "standard" looks like [this](https://tldp.org/LDP/abs/html/exitcodes.html) (many similar resources exist/document the desired behavior). That was the original motivation to never allow passing with `2..254`.

>     * wrong results -> return for now "1-255" or so, doesn't really matter.
I got this resolved over time, we consistently use `+1` for this case (`EXIT_FAILURE`).
> 
> Handling of the return value: When non-zero: when -1/"JIT Fails" check if the case is in the allow list.
> If yes -> mark allow, if no -> raise error.
That's implemented (and always was). I have to refresh allow-lists for eltwise, opreduce, and equation.

> When zero: check for allow listing: if case is in there -> raise error as this cannot be.
That's implemented and our default behavior for detecting "shaky" tests. if no -> mark success.

> When crash: raise Error/Crash (even if in allow list)
ACK. Requires changing back from `15` to `-1`.

alheinecke (commented):
done

PR #724, opened 2023-01-02
Refactor Eltwise Test

alheinecke:
This PR splits the big test scripts for unary and binary eltwise tests into many small ones based on TPP type and precision combination. Additionally the scripts are now generate during `make `which will make future maintenance much simpler.  The sample size of the random test samples can be also determined when running make such that all 1100 test can complete during a quick check in a couple of seconds.

@egeor : please conduct a carefully review, we have to make sure that I didn't accidentally removed some tests
@egeor : it seems on AVX2 there are issues with replicate_col_var for bf16/f16, can you please have look
@hfp : some allow list might need some updating
@all: I think we can merge if the first 2 item on the list are done. GVT3 doesn't need fixing as we have already branches for this. The merge of this PR is prerequisite such that we can start with SSE eltwise.


egeor (approved):
Looks good to me w.r.t. test generation scripts

PR #729, opened 2023-01-09
drop two-phase compilation targeting knc

hfp:
Dropping two-phase compilation targeting MIC architecture (KNC) affects the build system as well as the source code:

* Begin/end macros decorating symbols originating from external headers, and API decorators, e.g., `LIBXSMM_API`
* Build system (`Makefile`, `Makefile.inc`, and SpecFEM sample)


hfp:
Makefile internal variable: `NO_OFFLOAD_FLAG`.

PR #738, opened 2023-02-10
Added a driver for the bf16 3-way split of fp32 tensors via a TPP equation with DUMP operators.

kvoronin-intel:
Rationale: this would be a regression test for the recently fixed bug (see https://github.com/libxsmm/libxsmm/commit/7955cb14e8057525de20e7d09d8f911b4ca80f75) and a proxy for testing the new TPP if it appears for this split.

Tested locally with
  ./equation_bf16_x3_split_f32 17 31 17
  ./equation_bf16_x3_split_f32 17 31 25


samples/equation/equation_test/equation_bf16_x3_split_f32.sh
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/env bash
+
+source setup_eqn_tpp_prec_list.sh equation_bf16_x3_split_f32
+
+if [ -x "$(command -v python3)" ]; then
+  PYTHON=$(command -v python3)
+else
+  PYTHON=$(command -v python)
```
kvoronin-intel, line 8: Not sure if we want to add the new test or not but here I did the same as we have for other equations like splitSGD.

alheinecke (approved):

PR #741, opened 2023-03-04
Fix for REPLICATE_COL_VAR in BF16

ddkalamk:


alheinecke:
@egeor : it seem neither before or after the change our CI reacts, can you please double check?

egeor:
The culprit is here:

https://github.com/libxsmm/libxsmm/blob/main/samples/eltwise/kernel_test/generate_unary_simple_test_scripts.sh#L24

We allow non-F32 compute only for copy in the CI tests for unary simple. We should also allow this for the replicate col var (of course a top Dhiraj's commit)

alheinecke:
However, also ReLU & XoR should be tested for compute precision being low precision, I guess... and they all should be tested for FP8 at as well.

hfp:
Just FYI, our AArch64 based CI-pipeline is down ATM.

src/generator_mateltwise_aarch64.c
```diff
@@ -211,6 +211,7 @@ libxsmm_blasint libxsmm_generator_mateltwise_aarch64_valid_arch_precision( libxs
 
   if ((is_transform_tpp == 0) && (is_gather_scatter_tpp == 0) &&
       !((i_mateltwise_desc->operation == LIBXSMM_MELTW_OPERATION_UNARY) && (i_mateltwise_desc->param == LIBXSMM_MELTW_TYPE_UNARY_XOR)) &&
```
ddkalamk, line 213: BTW, aarch64 code path seems to miss check for ReLU as well.

egeor: On aarch64 we support relu only with F32 compute...

alheinecke:
yes... ReLU is a crazy mess... even on x86 it's only BF16 compute for EVEX encodings.... I'm right now prepping a better test here as well

PR #744, opened 2023-03-10
Adding fp32 ->bf16x2 / bf16x3 split tpp

alheinecke:


egeor (commented):
Looks good to me. The only open is integration with equations. Going with stack is obviously the easiest thing, the only thing that needs attention would be to setup in the struct param in stack the secondary offsets a la gathers/scatters/unpack.

alheinecke:
yes, this we can do later, also AVX2, SVE support is missing.

alheinecke:
SVE added, ready for merge

PR #745, opened 2023-03-10
Fix unary unpack to blocks/binary pack APIs and add CI tests

egeor:


samples/eltwise/kernel_test/generate_binary_test_scripts.sh
```diff
@@ -21,6 +21,11 @@ for PREC in 'F32_F32_F32_F32' 'BF16_BF16_BF16_BF16' 'F32_F32_BF16_F32' 'F32_BF16
         continue
       fi
 
+      # Binary pack tp blocks 2 x BF16 -> F32 is only possible for 1 prec combination
```
alheinecke, line 24: Why only BF16 and FP32 Compute, Compute should also be BF16 (as there is no compute, like copy). 
We also should test I16/I32 (as we do for transposes).
While BF16 and I16 are needed, I'm not sure about F16....

src/generator_matequation_regblocks_avx_avx512.c
```diff
@@ -609,7 +610,7 @@ void libxsmm_generator_mateqn_load_arg_to_2d_reg_block( libxsmm_generated_code*
     for (in = 0; in < i_n_blocking; in++) {
       for (im = 0; im < i_m_blocking; im++) {
         cur_vreg = i_start_vreg + in * i_m_blocking + im;
-        if ( arg_info[i_arg_id].dtype == LIBXSMM_DATATYPE_I16 ) {
+        if ( i_is_unpack_load > 0 ) {
```
alheinecke, line 613: This seems like a general knob, but assumes that input is 16bit and output/compute is 32bit given the load instruction is used. I feel we don't want to add a general flag for this to the very common load_reg_block routine....

src/generator_matequation_regblocks_avx_avx512.h
```diff
@@ -118,7 +117,8 @@ void libxsmm_generator_mateqn_load_arg_to_2d_reg_block( libxsmm_generated_code*
                                                  unsigned int                            i_n_blocking,
                                                  unsigned int                            i_mask_last_m_chunk,
                                                  unsigned int                            i_mask_reg,
-                                                 unsigned int                            i_skip_dtype_cvt );
+                                                 unsigned int                            i_skip_dtype_cvt,
+                                                 unsigned int                            i_is_unpack_load );
```
alheinecke, line 121: Is this specializing (which has various assumptions) a good idea? (I left also comment in the actual implementation)

src/generator_matequation_scratch_avx_avx512.c
```diff
@@ -489,6 +490,8 @@ void libxsmm_generator_matequation_set_output_in_stack_param_struct(libxsmm_gene
     } else {
       if (cur_node->info.u_op.type == LIBXSMM_MELTW_TYPE_UNARY_SCATTER) {
         fprintf( stderr, "The requested SCATTER operation can only be the head of the equation...\n" );
+      } else if (cur_node->info.u_op.type == LIBXSMM_MELTW_TYPE_UNARY_UNPACK_TO_BLOCKS) {
```
alheinecke, line 493: Similar to PACK -> 2X32BITUNPACK

samples/eltwise/kernel_test/generate_unary_simple_test_scripts.sh
```diff
@@ -26,6 +26,11 @@ for PREC in 'F32_F32_F32' 'BF16_BF16_BF16' 'BF16_BF16_F32' 'F32_BF16_F32' 'BF16_
         continue
       fi
 
+      # Unary unpack to blocks FP32 -> 2 x BF16 in only possible for 1 prec combination
+      if [[ ("$TYPE" == '42') && ("$PREC" != 'F32_BF16_F32') ]]; then
```
alheinecke, line 30: Similar to unpack, why not I16/I32 and also I16/BF16 as compute?

alheinecke (commented):
several minor issues and the bigger on the matequation with packflag for the load routine

samples/eltwise/run_test_a64fx.txt
```diff
@@ -110,62 +110,6 @@ binary_div_hf8_hf8_f32_f32_eqld
 binary_div_hf8_hf8_f32_f32_gtld
 binary_div_hf8_hf8_hf8_f32_eqld
 binary_div_hf8_hf8_hf8_f32_gtld
-binary_mul_bf16_bf16_bf16_f32_eqld
```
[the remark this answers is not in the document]
egeor: I think there were duplicate entries + they were not alphabetically sorted..

samples/eltwise/eltwise_binary_simple.c
```diff
@@ -158,6 +160,8 @@ void set_binarytype(unsigned int op, libxsmm_meltw_binary_type *type) {
     binary_type = LIBXSMM_MELTW_TYPE_BINARY_DIV;
   } else if ( op == MULADD_OP ) {
     binary_type = LIBXSMM_MELTW_TYPE_BINARY_MULADD;
+  } else if ( op == PACK_OP ) {
+    binary_type = LIBXSMM_MELTW_TYPE_BINARY_PACK;
```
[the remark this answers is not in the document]
egeor: sounds good

samples/eltwise/kernel_test/generate_binary_test_scripts.sh
egeor: For the binary pack the compute prec should be F32/I32 otherwise the vlen logic (i.e. least common denominator of input/compute) would be broken/incorrect and things will break lose later on... (or we have to special case for tis TPP which also sounds weird to override the generic logic). For BF16/I16/F16 input we can make it work trivially

src/generator_matequation_regblocks_avx_avx512.c
egeor: Correct, but the alternative without adding the knob would be to add a special function to be called which shares 99% code with the generic load... So adding a knob seemed more natural (albeit it is a specialization knob)

src/generator_matequation_scratch_avx_avx512.c
egeor: Yeah something like 2x16BITUNPACK I guess

samples/eltwise/kernel_test/generate_unary_simple_test_scripts.sh
egeor: For the unpack is fine, for pack not, see my other comment

src/generator_matequation_regblocks_avx_avx512.c
alheinecke: The ARM code doesn't need it. So let's rework this, we don't nee to duplicate the whole routine, it's just that the right information might not be available inside the routine (however, ARM code pass the skip CVT and BF16 check so I'm not sure why x86 cannot do the same). E.g. it would be better to pass the op to the routine and then decide internally what to do, than to make a global flag for a specific OP.

samples/eltwise/kernel_test/generate_binary_test_scripts.sh
alheinecke: I'm coming from this side: The operation doesn't do any compute and all other transform, data-reorder TPP ignore? the compute precesion and also here should only determine the vlen in/out and not compute precision. If the implementation of pack/unpack uses vlen based on the compute datatype it's a bug in itself. Where is this happening?

samples/eltwise/run_test_a64fx.txt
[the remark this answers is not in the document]
alheinecke: then please sort, such the diff gets down to what is actually added.

samples/eltwise/kernel_test/generate_unary_simple_test_scripts.sh
alheinecke: I feel even that PACK should be I16 -> I32/F32 and unpack should be F32/I32  -> I32. All our combination don't make sense. (e.g. when we split FP32 we don't get two BF16/FP16 values and we would pack/combine two F16/BF16 values we won't get a FP32 value).
The compute precision/datatype needs some thought... I feel right now it should be even ignored, perhaps we need to add this one ?!

samples/eltwise/kernel_test/generate_binary_test_scripts.sh
egeor: Here it is:
https://github.com/libxsmm/libxsmm/blob/main/src/generator_mateltwise_unary_binary_avx_avx512.c#L240

src/generator_matequation_regblocks_avx_avx512.c
egeor: For arm the flow is slightly different than x86 and some of the shebang is done in the store part (albeit with a specific function). Passing an OP and decide based on that is also fine with me (and indeed more generic).

samples/eltwise/kernel_test/generate_unary_simple_test_scripts.sh
egeor: Yeah indeed the precision is moot for such computations... It would make sense something like LIBXSMM_DATATYPE_16BIT/32BIT etc.. because even the I16 and I32 stuff mathematically is not correct...

samples/eltwise/kernel_test/generate_binary_test_scripts.sh
alheinecke: This needs to be fixed. There is reason why we have input, compute and output datatype width. Setting two to be globally the same is IMHO not the right thing to do.

src/generator_matequation_regblocks_avx_avx512.c
alheinecke: We should align ARM and x86 (so that in a possible future we can merge parts of the simpler code gen). Diverged stuff should only happen for good reason. E.g. if stuff is logically happening for store on the ARM that is not good either.
On a second thought why doesn't simply providing the old interface not work? We have a datatype in memory and we have datatype we want to do compute in, then we don't need a op in the load routine?

samples/eltwise/kernel_test/generate_unary_simple_test_scripts.sh
alheinecke: What is different between 16BIT value and I16? Of course the split etc. will changed the value, but the convention for such things is to call them integer types. Eg. real ISA has I16 shuffles which we can run on I32 data. In this particular case we separate/join FP32 into 2 I16. I don't see a reason to call this 16BIT instead of I16. So Ixx stands for xx bits. we might want to call it U16 (what is still another refactoring needed), as U16 is really just 16bits without 2 complement interpretation possibility.

src/generator_matequation_regblocks_avx_avx512.c
egeor: The old API merely does not have the compute dtype... it tries  always to cvt to f32 unless the relevant flag is used (which was introduced for the dump operator). Agreed that with compute dtype info as arg then it would also be complete and future proof with other compute dtypes

samples/eltwise/kernel_test/generate_unary_simple_test_scripts.sh
egeor: OK, let's go then with the common convention and support these ops only with F32->2xI16 and 2x16->F32

include/libxsmm_typedefs.h
```diff
@@ -100,9 +100,9 @@
   (LIBXSMM_DATATYPE_I64         == LIBXSMM_GETENUM_UNP(SRC)) ? LIBXSMM_DATATYPE_I64 : ( \
   (LIBXSMM_DATATYPE_U64         == LIBXSMM_GETENUM_UNP(SRC)) ? LIBXSMM_DATATYPE_I64 : ( \
   (LIBXSMM_DATATYPE_I32         == LIBXSMM_GETENUM_UNP(SRC)) ? LIBXSMM_DATATYPE_I32 : ( \
-  (LIBXSMM_DATATYPE_U32         == LIBXSMM_GETENUM_UNP(SRC)) ? LIBXSMM_DATATYPE_I32 : ( \
```
alheinecke, line 103: LIBXSMM_GETENUM_INP should be sign agnositc, so this needs to change back. 

(I64 and I8 still sign agnostic....)

samples/eltwise/eltwise_unary_simple.c
```diff
@@ -429,16 +451,26 @@ int test_unary_op( const libxsmm_blasint M, const libxsmm_blasint N, const libxs
...
   out       = (char*) libxsmm_aligned_malloc( LIBXSMM_TYPESIZE(dtype_out)*N_out*ldo, 64 );
   out_gold  = (char*) libxsmm_aligned_malloc( LIBXSMM_TYPESIZE(dtype_out)*N_out*ldo, 64 );
   _in       = in;
 
   /* init in */
-  init_random_matrix( dtype_in,  in,       1, ldi, N, 0 );
-  init_zero_matrix(   dtype_out, out,      1, ldo, N_out );
-  init_zero_matrix(   dtype_out, out_gold, 1, ldo, N_out );
+  if (unary_type == LIBXSMM_MELTW_TYPE_UNARY_UNZIP) {
+    init_random_matrix( dtype_out, in,       1, ldi*2, N, 0 );
```
alheinecke, line 465: Why setting N_out to 2x and then also 2xldo?

src/generator_mateltwise_sse_avx_avx512.c
```diff
@@ -276,10 +276,10 @@ void libxsmm_generator_mateltwise_update_micro_kernel_config_dtype_aluinstr( lib
     if ( (LIBXSMM_DATATYPE_F64 == LIBXSMM_GETENUM_INP( i_mateltwise_desc->datatype )) || (LIBXSMM_DATATYPE_I64 == LIBXSMM_GETENUM_INP( i_mateltwise_desc->datatype )) ) {
       io_micro_kernel_config->datatype_size_in = 8;
       io_micro_kernel_config->vmove_instruction_in = LIBXSMM_X86_INSTR_VMOVUPD;
-    } else if ( (LIBXSMM_DATATYPE_F32 == LIBXSMM_GETENUM_INP( i_mateltwise_desc->datatype )) || (LIBXSMM_DATATYPE_I32 == LIBXSMM_GETENUM_INP( i_mateltwise_desc->datatype )) ) {
+    } else if ( (LIBXSMM_DATATYPE_F32 == LIBXSMM_GETENUM_INP( i_mateltwise_desc->datatype )) || (LIBXSMM_DATATYPE_I32 == LIBXSMM_GETENUM_INP( i_mateltwise_desc->datatype )) || (LIBXSMM_DATATYPE_U32 == LIBXSMM_GETENUM_INP( i_mateltwise_desc->datatype ))  ) {
```
alheinecke, line 279: I guess here the wrong macro is used (hence the change in the macro). The right macro is LIBXSMM_GETENUM_UNP. This applies to the entire diff in the file.

src/generator_matequation_regblocks_avx_avx512.c
```diff
@@ -527,9 +527,9 @@ unsigned int libxsmm_generator_matequation_regblocks_vmove_instruction(libxsmm_d
   const int datatype = LIBXSMM_GETENUM_INP(dtype);
```
alheinecke, line 527: LIBXSMM_GETENUM_UNP

alheinecke (changes requested):
Some macro changes are needed.

alheinecke:
When can we get this landed?

samples/eltwise/run_test_aarch64.txt
```diff
@@ -1,3 +1,11 @@
+unary_decomp_fp32_to_bf16x2_f32_bf16_f32_eqld
```
alheinecke, line 1: as above

src/generator_mateltwise_sse_avx_avx512.c
```diff
@@ -306,10 +306,10 @@ void libxsmm_generator_mateltwise_update_micro_kernel_config_dtype_aluinstr( lib
...
-      } else if ( (LIBXSMM_DATATYPE_F32 == LIBXSMM_GETENUM_INP( i_mateltwise_desc->datatype1 )) || (LIBXSMM_DATATYPE_I32 == LIBXSMM_GETENUM_INP( i_mateltwise_desc->datatype1 )) ) {
+      } else if ( (LIBXSMM_DATATYPE_F32 == LIBXSMM_GETENUM_INP( i_mateltwise_desc->datatype1 )) || (LIBXSMM_DATATYPE_I32 == LIBXSMM_GETENUM_INP( i_mateltwise_desc->datatype1 )) || (LIBXSMM_DATATYPE_U32 == LIBXSMM_GETENUM_UNP( i_mateltwise_desc->datatype1 )) ) {
         io_micro_kernel_config->datatype_size_in1 = 4;
         io_micro_kernel_config->vmove_instruction_in1 = LIBXSMM_X86_INSTR_VMOVUPS;
-      } else if ( (LIBXSMM_DATATYPE_BF16 == LIBXSMM_GETENUM_INP( i_mateltwise_desc->datatype1 )) || (LIBXSMM_DATATYPE_I16 == LIBXSMM_GETENUM_INP( i_mateltwise_desc->datatype1 )) || (LIBXSMM_DATATYPE_F16 == LIBXSMM_GETENUM_INP( i_mateltwise_desc->datatype1 )) ) {
+      } else if ( (LIBXSMM_DATATYPE_BF16 == LIBXSMM_GETENUM_INP( i_mateltwise_desc->datatype1 )) || (LIBXSMM_DATATYPE_I16 == LIBXSMM_GETENUM_INP( i_mateltwise_desc->datatype1 )) || (LIBXSMM_DATATYPE_U16 == LIBXSMM_GETENUM_UNP( i_mateltwise_desc->datatype1 )) || (LIBXSMM_DATATYPE_F16 == LIBXSMM_GETENUM_INP( i_mateltwise_desc->datatype1 )) ) {
```
alheinecke, line 312: why are we using somethings "UNP" and then "UOP"

alheinecke: wouldn't it be better to switch to libxsmm_meltw_getenum_precision

alheinecke (changes requested):
minor nit

@ddkalamk  : I guess this will lead to some changes reqd in the extensions.

samples/eltwise/run_test_aarch64.txt
egeor: Yeah, this is due to some conflict when I merged, will fix

ddkalamk:
> minor nit
> 
> @ddkalamk : I guess this will lead to some changes reqd in the extensions.

I am not sure what exactly is changing that will impact extension code, to me it looks internal implementation changes, may be I am missing some API changes..

alheinecke:
existing TPP name changes and some API @egeor  has the details

src/generator_mateltwise_sse_avx_avx512.c
alheinecke: the other file is cleaned so I would mine if you do it here and clean up this UNP and UOP I even don't know what the difference is without looking out the macro definition.

samples/eltwise/run_test_aarch64.txt
alheinecke: great!

src/generator_matequation_regblocks_avx_avx512.c
alheinecke: I guess this was fixed

src/generator_mateltwise_sse_avx_avx512.c
alheinecke: Also fixed?

samples/eltwise/eltwise_unary_simple.c
alheinecke: Is this still open or aligned with the split?

include/libxsmm_typedefs.h
alheinecke: fixed?

samples/eltwise/eltwise_binary_simple.c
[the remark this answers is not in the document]
alheinecke: Is this done?

samples/eltwise/kernel_test/generate_binary_test_scripts.sh
alheinecke: This is fixed?

src/generator_matequation_regblocks_avx_avx512.c
alheinecke: This is addressed?

src/generator_matequation_scratch_avx_avx512.c
alheinecke: Fixed?

samples/eltwise/eltwise_binary_simple.c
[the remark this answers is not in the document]
egeor: yes

samples/eltwise/kernel_test/generate_binary_test_scripts.sh
egeor: yes

src/generator_matequation_regblocks_avx_avx512.c
egeor: yes

src/generator_matequation_scratch_avx_avx512.c
egeor: yes

include/libxsmm_typedefs.h
egeor: yes

samples/eltwise/eltwise_unary_simple.c
egeor: I didn't change this

src/generator_mateltwise_sse_avx_avx512.c
egeor: yes

src/generator_matequation_regblocks_avx_avx512.c
egeor: yes

src/generator_matequation_regblocks_avx_avx512.h
alheinecke: I think this was also fixed?

alheinecke (commented):
more comments

samples/eltwise/eltwise_unary_simple.c
alheinecke: In line 455 you added the code which is aligned with the split. this seems to be now not necessary

alheinecke (commented):
more comments

src/generator_matequation_regblocks_avx_avx512.h
egeor: yes

samples/eltwise/eltwise_unary_simple.c
egeor: Correct, forgot I did this, will cleanup

egeor:
> > minor nit
> > @ddkalamk : I guess this will lead to some changes reqd in the extensions.
> 
> I am not sure what exactly is changing that will impact extension code, to me it looks internal implementation changes, may be I am missing some API changes..

Basically the unpack is renamed to LIBXSMM_MELTW_TYPE_UNARY_UNZIP (with input dtype U32 and outputs U16) and the PACK is renamed to LIBXSMM_MELTW_TYPE_BINARY_ZIP (with inputs U16 and output U32).

src/generator_mateltwise_sse_avx_avx512.c
egeor: OK switched to libxsmm_meltw_getenum_precision

PR #751, opened 2023-03-24
refactor utils

hfp:
Reduce the "visible" API (surface) and introduce utilities. This can evolve into a separate sub-library (`libxsmmutils`), but not necessarily (if the utilities are still used internally in the main library).


alheinecke (changes requested):
This PR need to be broken down into multiple pieces, I don't see how it's possible to merge it in its current form. The goal was to reduce entanglement, but now I see we require users to add more includes. That's not good. I will put more detailed description with issues as a larger comment.

alheinecke:
The initial comment by @hfp should be the guiding design principle: This can evolve into a separate sub-library (`libxsmmutils`). There first round of changes/refactors should pick the low-hanging fruits without creating any major user-level changes, like requiring additional includes when attempting to use base library functionality. Right now this is not the case, in order to use all TPPs and some of the datatypes at lease 3 new includes (rng, math, lpquant) are needed. However, the PR is super-help to highlight these entanglements and part of the PR ready for merge (see below), but these are IMHO blocked due to the newly created entanglement. 

Proposal: Let's split the PR into several PRs. Here is a little assessment from my side after studying the PR and trying to come up with a list of PRs with work towards a separate sub-library:

`libxsmm_x86_intriniscs.h` : medium, there are some intrinsics needed in the main library (for hashing), but 95% of the code is unused. Perhaps a slit into the absolute minimal internal part and the nice extensions. 

`libxsmm_lpflt_quant.h`: hard, LIBXSMM introduces new floating point datatypes (bf16, fp16, bf8, hf8, fixpoint) which are not part of C. This file offers scalar converts to and from these types and also large buffer convert routines. Perhaps the best is to leave the file in the main library but only keeping the scalar convert routines. The buffer converts could move into utils. The scalar converts could merge with `libxsmm_math.h`

`libxsmm_math.h`: easy, after grepping for the usage -> it should remain with the main-library, it's comparable to the malloc functionality which is part of the main library.

`libxsmm_memory.h`: easy, after grepping for the usage -> it should remain with the main-library, it's comparable to the malloc functionality which is part of the main library.

`libxsmm_mhd,h`: easy, can be part of the utils library as there is basically no entanglement, similar to many of the compound intrinsisc functions discussed above.

`libxsmm_rng.h`: hard, all state related stuff must be part of the main-library as otherwise PRNG, Dropout, stochastic round TPP won't work anymore. The scalar F32,I32,F64 random number gens are super helpful but could move into either `libxsmm_math.h` or an external utils library.

`libxsmm_sync.h`: hard, all the macros are needed for the code registry etc. should be part of the main library. the `libxsmm_barrier` could move into an external utils library.

`libxsmm_timer.h`: easy, when having an external library for the compound intrinsiscs, mhd, `libxsmm_barrier`, the `libxsmm_timer` fits right into it.

`libxsmm_utils.h`: easy this is parallel gemm, batch gemms etc. 

We would need 4-5 PRs:
-  "easy" items could land right away: `libxsmm_utils.h`, `libxsmm_timer.h`, `libxsmm_mhd,h`
- split of `libxsmm_x86_intrinsiscs`
- split of `libxsmm_sync.h`
- split of `libxsmm_lpflt_quant.h`
- moving `libxsmm_rng.h` into  `libxsmm_math.h`

With this `libxsmm` would have the main critical components to write code against TPPs and being thread save, all with one include and one library. `libxsmmutils` become collection of truly "on-top" functionality which can help with parallel code, performance debugging etc.

hfp:
The `libxsmm_timer.h` is still a bit involved as we do some loop-calibration to get real/wall time.

hfp:
Work is mostly complete (except for `libxsmm_x86_intrinsiscs.h` and `libxsmm_timer.h`). Everything so far was depending on everything else like even breaking out the barrier was depending on synchronization primitives and other stuff. I will now start breaking the changes artificially into separate PRs as much as possible/sensible.

The timer and intrinsics work will remain as truly separate work/PRs (again a big package as we still have tons of code with target attributes and manual dispatch).

hfp:
>  require users to add more includes

By design, users need to opt-in for utility functionality, i.e., `libxsmm_utils.h` is separate from `libxsmm.h`. Of course, individual utilities (with potentially own header files under `include/utils`) are accessible by just including `libxsmm_utils.h`.

Issue #753, opened 2023-03-24
GEMM tester (xgemm) has convoluted format specification, not all variants are tested


alheinecke:
Right now we the xgemm driver has only 2 flags for formats (a trans and b trans). However with latest additions we have various VNNI factors and formats.

We therefore we need at least 6 flags: a-trans, b-trans, c-trans, a-vnni, b-vnni, c-vnni.

Additionally we have the problem that platforms like GVT3 have support for bfdot and bfmmla which are selected right now through env variables for such systems. The CI should read during script generation the VNNI factor.

alheinecke:
This issues has been addressed in all aspects which are not related to GVT3 and support bfdot and bfmmla.

alheinecke:
bfmmla is handled via btrans/bvnni.

PR #754, opened 2023-03-27
SRF & GNR initial support

alheinecke:
Added all relevant SRF and GNR instructions
Enhanced LIBXSMM CPUID & Dispatch to handle SRF like ADL and GNR like SPR
Added BF16 GEMM that makes use of AVX-NE-CONVERT instructions
int8 (all sign combos for AVX2) and FP16 GEMM (AVX2 & AMX) are not included in this PR and can only be done after #753 has been addressed.


hfp (approved):
Looks good to me!

PR #755, opened 2023-03-31
Refactored libxsmm_mmfunction functor to rely on v2-dispatch

hfp:


alheinecke:
Let's review PRs in future. This on is hopefully fine as it will not create issues with private branches, but it's always good to review.

PR #757, opened 2023-04-10
Started RISCV support

siddharthrai:
These changes contain encoding support for RISCV RV64I instructions


hfp:
I wonder why test results are still not shown on this page (or actually tested on our infra). I guess it's because your PR is prefixed automatically with your GitHub-username and then does not match our branches to be tested. I will take a look on relaxing/adjusting our patterns.

Btw, your code can perfectly merge `main_stable` branch and I would perhaps do this from time to time, but let's wait for me relaxing our branch patterns...

alheinecke (changes requested):
Can you please remove the files "generator_riscv_instrutions.c/h" they are now committed 2x there also added "generator_rv64_insturctions.c/h" we only want to keep the later on.

PR #758, opened 2023-04-28
Fix bf8 stoch round

alheinecke:


alheinecke:
@nkmellem @ddkalamk @egeor : while this PR is now passing CI, we still need to limit / overwrite internally for the eltwise TPPs when stochastic rounding makes sense:
Can we agree on the following:
- when input is BF8 and the TPP is memory movement -> a set stochastic round-flag is ignored (that incl. copy, replicat col vars and transforms)
- for XOR the stochastic round flag is ignored
- What about equations, stochastic rounding doesn't seem to enabled right now @egeor : what would be needed
- What about the reduce TPP?

PR #786, opened 2023-06-19
Removed all occurrences of JUMP_LABEL_TRACKER_MALLOC

hfp:
* The tracker was free'd at scope's end, which is exactly what the stack based tracker serves.
* If there was a case for malloc not being free in the same scope, it was a resources leak.
* This code caused static analysis (SA) to claim resource leakage, e.g., in case of an error (LIBXSMM_HANDLE_ERROR+return).
* The issue was raised by recently added BCSC code not covering all error paths by freeing memory (pending SA issues).
* Removed all JUMP_LABEL_TRACKER_MALLOC related code and rely on pointing to stack based loop-label tracker.
* Note: some BE-functions still get the tracker as well as the microkernel-config as part of their arguments.
* TODO: let libxsmm_micro_kernel_config embed tracker by-value, not by-pointer?

### DIFF
--- a/src/generator_common.h
+++ b/src/generator_common.h
@@ -1477,6 +1477,7 @@ LIBXSMM_EXTERN_C typedef struct libxsmm_micro_kernel_config {
   /* Auxiliary fields to propagate kernel info */
   unsigned int m_remainder;
   unsigned int br_loop_index;
+  /* TODO: should be by-value, not by-pointer? */
   libxsmm_jump_label_tracker *p_jump_label_tracker;
   unsigned int loop_label_id;
   unsigned int k_amx_microkernel;
--- a/src/generator_gemm_amx.c
+++ b/src/generator_gemm_amx.c
@@ -18,10 +18,6 @@
 #include "generator_mateltwise_unary_binary_avx_avx512.h"
 #include "generator_mateltwise_transform_common.h"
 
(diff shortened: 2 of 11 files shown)


hfp:
As a side-effect, this PR eliminates the recently introduced Coverity issues. Some code paths handling errors did not free memory used for scope-local loop-label tracker.

hfp:
> For now I will not merge such clean-ups as we branches open, need to first check all of them.

ACK. Though, new BCSC code brought up some issues in Coverity and the objective was not to add more/wrong code to fix it. This cleans it up however, and mostly removes lines whereas fixing the SA issues the usual way (like done for some of the return paths) otherwise adds more such bogus code.

alheinecke:
The problem are issues like #789 from all these refactor and this minimal touch ups. We have tight deliveries on features and now I have to work with local branches etc., simply because of behavior changes, minor merge conflicts which are fixed in-correctly. That's the whole purpose of main-stable, not doing such things and providing a stable development branch for all the features are in flight. With all of these back-ports we are creating slow-downs on the the development side of features, e.g. my solution right now is not to merge main_stable with #785 into feature branches as it not clear what else changes with #785. 

#785 was also merged without review, so such things could be avoidable. Maybe it's best to just revert #785 on main_stable and only merge necessarily things.

hfp:
This can wait until merging `main` and `main_stable`.

PR #788, opened 2023-06-19
Fix an issue with zero rows with FsSpMDM on NEON.

FreddieWitherden:


hfp:
I will white-list you for our CI, since tests did not pick your PR (I will merge and test). Thank you for this quick fix!

PR #795, opened 2023-06-29
(BR)GEMM variants i8bf16bf16/i8bf16f32 for x86 (>avx2)

egeor:
This PR introduces the (BR)GEMM variants i8bf16bf16/i8bf16f32 for x86 (>avx2)


hfp:
CI-wise this seems to be good to merge. Not judging the actual code or providing a review at this point. If this PR merges just yet into `main_stable`, we should aim to get rid of `main` and `main_stable` differences even sooner. This PR touches a lot of code in the backend (causing potential conflicts for pending work?). In fact, it might be an option to merge into `main` rather than `main_stable`. If so, I can assist.

alheinecke:
we still have several branches with changes open vs. main_stable. Over the next 3-4 weeks main_stable should have all of them merged, then main merge can be planned

src/generator_gemm_avx512_microkernel.c
```diff
@@ -59,13 +69,89 @@ void libxsmm_generator_gemm_avx512_kloop_kernel( libxsmm_generated_code*
     }
 
     for ( l_k = 0; l_k < i_k_blocking; l_k += l_k_pack_factor) {
+      if (l_is_Ai8_Bbf16_gemm > 0) {
```
alheinecke, line 72: This is very convoluted logic as it's even overwriting the same variable. I think we need to find a better & cleaner solution.

alheinecke (changes requested):
I don't understand why this is_last_k is needed. It seems create a lot of Spaghetti code and I would like us to step back and think if we can do this easier.

alheinecke (approved):

PR #797, opened 2023-07-07
(BR)GEMM variants bf8f16/i4f16 for x86 (>avx2)

egeor:
This PR introduces (BR)GEMM variants bf8f16/i4f16 for x86 (>avx2)


samples/xgemm/gemm_kernel.c
```diff
@@ -20,6 +20,8 @@
 #define RELU_BITMASK    2
 #define SIGMOID         3
 
+unsigned int l_is_Ai4Bf16_gemm = 0;
```
alheinecke, line 23: do we need a global variable?

alheinecke: I think we can just a field into the struct?

src/libxsmm_main.c
```diff
@@ -1836,6 +1836,18 @@ LIBXSMM_API_INLINE const char* libxsmm_get_gemm_typename(const unsigned char* da
       {
         return "i8f16f32";
       }
+      else if (LIBXSMM_DATATYPE_BF8 == LIBXSMM_GEMM_GETENUM_A_PREC(datatype) &&
```
alheinecke, line 1839: I think some print change would be need for the int4, otherwise we have non-unique code dumps.

src/generator_gemm_sse_avx_avx2_avx512.c
```diff
@@ -136,6 +136,10 @@ LIBXSMM_API_INTERN void libxsmm_generator_gemm_sse_avx_avx2_avx512_kernel( libxs
   const char *const env_bf8_gemm_via_stack_alloc_tensors = getenv("LIBXSMM_BF8_GEMM_VIA_STACK");
   int bf8_gemm_via_stack_alloc_tensors = 0;
   int hf8_gemm_via_stack_alloc_tensors = 0;
+  unsigned int l_is_Ai4_Bf16_gemm = (((i_xgemm_desc->flags & LIBXSMM_GEMM_FLAG_INTERPRETE_A_AS_INT4_VNNI2) > 0) &&
```
alheinecke, line 139: is additionally an arch check needed?

alheinecke (changes requested):
minor nits

samples/xgemm/gemm_kernel.c
egeor: The reason I didn't do it is that some API changes would be needed for the init matrix (i.e. pass the gemm struct as argument). If you are fine with this, we can get rid of the global

src/libxsmm_main.c
egeor: Will fix

src/generator_gemm_sse_avx_avx2_avx512.c
egeor: I think it taken care of in earlier part

samples/xgemm/gemm_kernel.c
alheinecke: Let's fix the API then as the struct has VNNI flags etc. I feel that flag also belongs there.

src/libxsmm_main.c
alheinecke: great, will resolve and wait for the push.

egeor:
This should be good now

alheinecke (approved):

PR #807, opened 2023-08-16
Backported frontend/implementation/api updates

hfp:


alheinecke (commented):
What is the reason for these frontend  changes?

hfp:
> What is the reason for these frontend changes?

* ABI spec is outdated on `main_stable` (`scripts/tool_check_abi.sh` would produce changes already even without this PR) like functions are not mentioned which are public (some of the `_v2` functions are removed on `main` already like `libxsmm_dmmdispatch_v2` and `libxsmm_smmdispatch_v2`).

* Implementation wise, there are some corrections for corner cases previously untested. Otherwise it's about clarifying code comments on existing functions. The [src/libxsmm_matrixeqn.c](https://github.com/libxsmm/libxsmm/pull/807/files#diff-1f8e939305cbc2182abb4c48772783690f9a4dd2844756e5d2dffd6d71089e48) change is a bugfix.

* The changes about coprime/2 and shuffle have no impact for existing code and are related to own research.

alheinecke:
Thanks, can we please for now limit the PR to:

ABI spec is outdated on main_stable (scripts/tool_check_abi.sh would produce changes already even without this PR) like functions are not mentioned which are public (some of the _v2 functions are removed on main already like libxsmm_dmmdispatch_v2 and libxsmm_smmdispatch_v2).

Implementation wise, there are some corrections for corner cases previously untested. Otherwise it's about clarifying code comments on existing functions. The [src/libxsmm_matrixeqn.c](https://github.com/libxsmm/libxsmm/pull/807/files#diff-1f8e939305cbc2182abb4c48772783690f9a4dd2844756e5d2dffd6d71089e48) change is a bugfix.

For all the other changes in main, I need to find to some to review how we can merge most of them and move main to a feature branch which has such private research features which we then can merge afterward.  But that at least would allow us the merge some refactor now, make main_stable main and we can merge into feature branches. The remainder (hopefully small) would live in a feature branch for now.

hfp:
The PR is now narrowed to the item you selected.

alheinecke (approved):

Issue #790, opened 2023-06-19
Why did Error-Bounds change with #785


alheinecke:
With #785 some error bounds (RCP on AARCH_V81, but not SVE256 ) where significantly changed, why is this, I cannot follow the conditions here. ASIMD and SVE should have identical flows through the tests.

https://github.com/libxsmm/libxsmm/blob/main_stable/samples/eltwise/eltwise_unary_simple.c#L536  vs. https://github.com/libxsmm/libxsmm/blob/aa2c13a8f287f7931b12621a162d5dacffa92b8e/samples/eltwise/eltwise_unary_simple.c#L541

alheinecke:
I understand there was some nicefying rewrite. However, my problem is much bigger: why did the error bounds change to "50" from "0.002" or so. "50" is clearly wrong as even for SVE it still seems to be old one. 

Can we please revise to error bars smaller or equal than what was there before. If this is not possible. Let's revert #785 and take it step by step.

alheinecke:
We can see that error bounds get loser and 0.1 -> 0.2, and the RCP if logic is not clear to me and the error bounds are completely off.

I haven't even looked at other tests like binary, relu etc.

But these changes are for sure nothing we can just merge.

hfp:
I will take a deeper look at unary simple and reciprocals and then open issues as needed (reproducer), and/or lower bounds such that CI turns red.

alheinecke:
any update on this?

hfp:
No update yet. This issue remains actionable and will get its attention!

alheinecke:
This issue is closed

PR #808, opened 2023-08-17
Issue #790: restore error bounds/logic

hfp:
* It can be assumed libxsmm_matdiff_epsilon mostly impacted by normf_rel (old/new error bounds applicable before/after).
* This keeps calculating error based on libxsmm_matdiff_epsilon (rather than normf_rel).
* Any tighter error bounds introduced by #785 are kept (tighter is better).


hfp:
I rebased on top of `main_stable`, this PR is ready also.

alheinecke (approved):

PR #810, opened 2023-08-22
Feature add min col idx reduce

egeor:
This PR adds the reduce min cols idx TPP, and relevant CI tests. For now this uses the opreduce TPP code gen, soon this will be deprecated with a future PR


include/libxsmm_typedefs.h
```diff
@@ -268,13 +270,13 @@ typedef enum libxsmm_meltw_unary_flags {
...
   LIBXSMM_MELTW_FLAG_UNARY_IDX_SIZE_4BYTES    = 128,
   LIBXSMM_MELTW_FLAG_UNARY_IDX_SIZE_8BYTES    = 256,
-  LIBXSMM_MELTW_FLAG_UNARY_GS_ROWS            = 512,
-  LIBXSMM_MELTW_FLAG_UNARY_GS_COLS            = 1024,
-  LIBXSMM_MELTW_FLAG_UNARY_GS_OFFS            = 2048,
-  LIBXSMM_MELTW_FLAG_UNARY_REDUCE_NEG_INF_ACC = 4096,
-  LIBXSMM_MELTW_FLAG_UNARY_REDUCE_RECORD_ARGOP= 8192,
-  LIBXSMM_MELTW_FLAG_UNARY_REDUCE_NO_PREFETCH = 16384,
-  LIBXSMM_MELTW_FLAG_UNARY_STOCHASTIC_ROUND   = 32768
+  LIBXSMM_MELTW_FLAG_UNARY_REDUCE_INF_ACC     = 512,
```
alheinecke, line 273: Do I get it right, that based on selected OP (min, max) the inf (pos, neg) is used?

alheinecke (commented):
Only a small question on visible frontend flag change. I would have expected some changes in some drivers/testers.

include/libxsmm_typedefs.h
egeor: yep

alheinecke:
Ah... can you please also a PR to here:
https://github.com/libxsmm/libxsmm-dnn/blob/main_stable/src/libxsmm_dnn_pool.c#L71

include/libxsmm_typedefs.h
alheinecke: thanks.

alheinecke (approved):

PR #809, opened 2023-08-20
Remove Support for KNL and KNM

alheinecke:
This is a long standing issue which I had started a long time ago in private branch. Background: The Knights product line was discontinued in 2019. We are running into issues to test these code paths efficiently. 
The biggest change is the collapse/remove of the AVX512_COMMON code path.

PR #812, opened 2023-08-23
Backported CI environment updates, build system updates, and minor fixes/refactor

hfp:
* Makefile.inc: allow to keep build state (STATEFLT=0). Permit deletion of PREFIX location if CLEAN=1 is given.
* Makefile/install: Support removing an existing PREFIX when "make PREFIX=... install" (CLEAN=1).
* Makefile: removed cumbersome dependency treating libxsmm_hash.c as header/include file.
* Adjusted sample code for portability, use signed integer for OpenMP-loop. Added VS project.
* Fixed libxsmm_shuffle (and refactored into libxsmm_shuffle and libxsmm_shuffle2).
* CI: removed "clxap" and "clxaptrb" (similar to or complements #809).
* Enable header-only installation for out-of-tree builds.
* Improved/simplified syntax (libxsmm_source.sh).
* Moved libxsmm_matdiff.h into source directory.


hfp:
CI-wise, this removes reference to "clxap" and "clxaptrb". Buildsystem wise, this now installs all bits when performing out-of-tree builds, and there is an option for "clean install" (TPP-mlir is in need like installing unclean can leave files deleted/renamed from repo). Added new sample code to VS and adjusted for compatibility (I may fix potential false sharing as exercise later on for malloc'd TLS).

samples/deeplearning/vnni_pack/vnni_pack.c
```diff
@@ -34,19 +35,21 @@ void pack_c(const libxsmm_bfloat16 *src, libxsmm_bfloat16 *dst, const unsigned i
...
-  unsigned int kBlocks = K/bk;
-  unsigned int cBlocks = C/bc;
-  LIBXSMM_VLA_DECL(2, libxsmm_bfloat16, real_src, src, K);
+  const unsigned int cBlocks = C/bc;
+  const int kBlocks = K/bk;
+  unsigned int k2, c1, c2;
+  int k1;
+  LIBXSMM_VLA_DECL(2, const libxsmm_bfloat16, real_src, src, K);
   LIBXSMM_VLA_DECL(5, libxsmm_bfloat16, real_dst, dst, cBlocks, bc/vnni_pack, bk, vnni_pack);
+  libxsmm_bfloat16 *const tmp = (libxsmm_bfloat16*)malloc(sizeof(libxsmm_bfloat16) * bc * bk * kBlocks);
```
alheinecke, line 44: This is a bug, the temp buffer needs to be inside the parallel region and stack allocated for performance, please revert these changes.

samples/deeplearning/vnni_pack/vnni_pack.c
```diff
@@ -82,14 +85,15 @@ void pack_tpp_normtovnni(const libxsmm_bfloat16 *src, libxsmm_bfloat16 *dst, con
 }
 
 void unpack_c(const libxsmm_bfloat16 *src, libxsmm_bfloat16 *dst, const unsigned int C, const unsigned int K, const unsigned int bc, const unsigned int bk, const unsigned int vnni_pack) {
-  unsigned int k1, k2, c1, c2;
-  unsigned int kBlocks = K/bk;
-  unsigned int cBlocks = C/bc;
+  const unsigned int cBlocks = C/bc;
+  const int kBlocks = K/bk;
+  unsigned int k2, c1, c2;
+  int k1;
```
alheinecke, line 91: why k1 signed and c1 not... this is weird we should use the same integer type for all loops.

samples/deeplearning/vnni_pack/vnni_pack.c
```diff
@@ -34,19 +35,21 @@ void pack_c(const libxsmm_bfloat16 *src, libxsmm_bfloat16 *dst, const unsigned i
...
 
 void pack_tpp_identity(const libxsmm_bfloat16 *src, libxsmm_bfloat16 *dst, const unsigned int C, const unsigned int K, const unsigned int bc, const unsigned int bk, const unsigned int vnni_pack, libxsmm_meltwfunction_unary kernel) {
-  unsigned int k1, k2, c1, c2;
-  unsigned int kBlocks = K/bk;
-  unsigned int cBlocks = C/bc;
-  LIBXSMM_VLA_DECL(2, libxsmm_bfloat16, real_src, src, K);
+  const unsigned int cBlocks = C/bc;
+  const int kBlocks = K/bk;
+  unsigned int k2, c1, c2;
+  int k1;
```
alheinecke, line 41: why k1 signed and c1 not... this is weird we should use the same integer type for all loops.

samples/deeplearning/vnni_pack/vnni_pack.c
```diff
@@ -12,14 +12,15 @@
...
 void pack_c(const libxsmm_bfloat16 *src, libxsmm_bfloat16 *dst, const unsigned int C, const unsigned int K, const unsigned int bc, const unsigned int bk, const unsigned int vnni_pack) {
-  unsigned int k1, k2, c1, c2;
-  unsigned int kBlocks = K/bk;
-  unsigned int cBlocks = C/bc;
-  LIBXSMM_VLA_DECL(2, libxsmm_bfloat16, real_src, src, K);
+  const unsigned int cBlocks = C/bc;
+  const int kBlocks = K/bk;
+  unsigned int k2, c1, c2;
+  int k1;
+  LIBXSMM_VLA_DECL(2, const libxsmm_bfloat16, real_src, src, K);
```
alheinecke, line 19: why k1 signed and c1 not... this is weird we should use the same integer type for all loops.

include/libxsmm_math.h
```diff
@@ -76,7 +76,9 @@ LIBXSMM_API size_t libxsmm_lcm(size_t a, size_t b);
  */
 LIBXSMM_API int libxsmm_primes_u32(unsigned int num, unsigned int num_factors_n32[]);
 
-/** Calculate co-prime number <= n/2 (libxsmm_coprime2(0|1) == 0). */
+/** Co-prime R of N such that R <= MinCo (libxsmm_coprime2(0|1) == 0). */
+LIBXSMM_API size_t libxsmm_coprime(size_t n, size_t minco);
```
alheinecke, line 80: We decide to keep this out for now (see the older PR), can you please remove this?

alheinecke (changes requested):
I see several issues with your rewrites, the changes to const. are fine though

samples/deeplearning/vnni_pack/vnni_pack.c
hfp: Ok, then such/previous code uses VLA like bounds of this stack-allocated array are defined at runtime (does not compile with our baseline). Alternatively, the slots per thread could be aligned/sized to multiples of the cacheline size. Raising baseline to C99 would be an option too, but then the project does compile under Windows.

samples/deeplearning/vnni_pack/vnni_pack.c
hfp: True. The suggested code avoids warnings about comparing two different signs and uses signed integers for the OpenMP loop. The latter is a prerequsite for correct OMP (unsigned loop induction variables are not accepted unless lazy compiler/settings). Alternatively, I can make everything signed and take the hit from smaller range like incoming data is unsigned. The latter does not matter too much for the outer/parallelized loop since the number of blocks is more likely within bounds.

samples/deeplearning/vnni_pack/vnni_pack.c
hfp: See above explanation.

samples/deeplearning/vnni_pack/vnni_pack.c
hfp: See above explanation.

include/libxsmm_math.h
hfp: It's refactored and the implementation is bugfixed this way. Though coprime2 function is just present since long time. Refactoring and improved test are not exactly a feature or new API.

samples/deeplearning/vnni_pack/vnni_pack.c
alheinecke: We can also make the stack allocation 128*128*4 = 64K and check bc / bk. However, the buffer needs to be on the stack such that we can compare with the TPP-MLIR which will do the same.

Windows is right now irrelevant. LIBXSMM doesn't implement the right calling convention and may give wrong results under windows. So let's not worry about windows.

samples/deeplearning/vnni_pack/vnni_pack.c
alheinecke: The OpenMP uses collapse, so both loop would need this. It seems you are testing with a broken compiler and I would not use this compiler anymore. 

Making everything signed, but currently the code is not in good shape with mixing signed / unsigned.

include/libxsmm_math.h
alheinecke: OK, I'm getting confused by all the refactor. Let's try for now to do only necessary refactors and focus on getting features in.

alheinecke:
This PR needs updates as I had to HOTFIX the CI portion directly as the nodes are gone.

alheinecke:
Let's break this into multiple commits as so much has changed: a) Make system changes b) CI changes c) rest.

PR #814, opened 2023-08-28
Fixed mantissa threshold for hf8 overflow convert

egeor:


src/libxsmm_lpflt_quant.c
```diff
@@ -576,7 +576,7 @@ LIBXSMM_API void libxsmm_rne_convert_fp32_hf8(const float* in, libxsmm_hfloat8*
       m = 0x7;
     /* overflow --> make it NaN */
     } else if ( (e_f32 > (f32_bias - f8_bias + 15)) ||
-                ((e_f32 == (f32_bias - f8_bias + 15)) && ( m_f32 > 0x600000))) {
+                ((e_f32 == (f32_bias - f8_bias + 15)) && ( m_f32 > 0x680000))) {
```
alheinecke, line 579: Why is a similar fix not needed for code gen?

alheinecke (commented):
small question.

src/libxsmm_lpflt_quant.c
egeor: We had decided to use only the routine with f32->f16 and f16->hf8 path for the codegen (this is is just there for future reference)

alheinecke: ACK.

alheinecke (approved):

alheinecke:
btw, do we also need changes to our bias conversion routines?

egeor:
> btw, do we also need changes to our bias conversion routines?

we do not have hf8 cvt with bias in libxsmm...

Issue #819, opened 2023-09-14
Windows support


jgong5:
Does libxsmm support Windows?

alheinecke:
yes, but we don't validate / CI on Windows. Do you run into issues? There might be an issue with some callee-save registers. I'm tracking this as a low-prio issue right now.

jgong5:
> yes, but we don't validate / CI on Windows. Do you run into issues? There might be an issue with some callee-save registers. I'm tracking this as a low-prio issue right now.

We are enabling IPEX on Windows but we haven't enabled the tpp/libxsmm related features in IPEX right now. There are some compilation issues we need to look into. Just wanted to confirm its support here. May I know if you will add this into you validation? Also, may I know what are impacted with the callee-save register issue?

jgong5:
cc @xuhancn

hfp:
> We are enabling IPEX on Windows but we haven't enabled the tpp/libxsmm related features in IPEX right now. There are some compilation issues we need to look into. Just wanted to confirm its support here.

> May I know if you will add this into you validation?

It is possible of course, but low priority on our side. In principle, I prepared our Makefile based build system compatible with Windows (spaces in path, etc). I am also keeping VS projects up-to-date at least wrt latest VS. The latter can be automated too, and there are a few Open Source CI-provider operating on Windows. Beside of the latter, we can likely get cloud support to do so on our own.

> Also, may I know what are impacted with the callee-save register issue?

I did this work a very long time ago, and our regmapping was not updated since then even when we changed our calling convention/interface for kernels. Assume it may work by luck.

alheinecke:
PR #822 fixes this issue, once merged we can close this issue

PR #822, opened 2023-10-07
Finalize Windows support

alheinecke:
This PR admends 98be791e8a4e996e017edb37347fb9bb4c4c5dc9 by adding support for saving also xmm6-xmm15 which are callee save on Win x64. 

This also addresses #819.


egeor (approved):
Looks good to me

PR #828, opened 2023-11-13
Fix Const-Cast warnings

alheinecke:
fixing warnings when compiling LIBXSMM with -Wcast-qual:
    - fixed dropping of const qual where it could be avoided without perf
      loss
    - selectively disabling the warning when calling into JIT'ed ASM code


rengolin:
You still have some warnings on the clang build. Without `-Werror` you will have to dig them out every build...

rengolin:
Though now I see this PR only fixes const casts. The warnings on the build are across different types (the worse kind), so fair. I'll let @egeor approve.

rengolin (approved):
Just spoke offline, the warnings are on tests and that's why tests, so we don't worry.

PR #837, opened 2023-12-05
Resolved cast-qual and other warnings

hfp:
* Makefile/inc
  - Only apply certain compiler flags if available with compiler in question.
  - Avoid duplicated warnings flags.
  - Fallback for LIBM and LIBC.

* Warnings
  - Introduced LIBXSMM_ASSIGN_VALUE, and resolved cast-qual warnings.
  - Avoid GCC-specific directive and rely on family of LIBXSMM_PRAGMA_DIAG_* macros.
  - Utilize LIBXSMM_PRAGMA_DIAG family of directives.
  - Avoid warnings about unused function.
  - Fixed numerous compilation warnings.
  - Fixed dropping const-qualifier.

* Macros/infra
  - Improved LIBXSMM_TYPEINFO to support FP-, and UI-info. Made LIBXSMM_TYPEINFO_FP, and LIBXSMM_TYPEINFO_UI separately available.
  - Make diagnostic-control available early, i.e., include headers with warnings suppressed.
  - Adjusted condition for older Clang (LIBXSMM_PRAGMA_DIAG_OFF_FADDRESS).
  - Made LIBXSMM_PRAGMA available more broadly.
  - Enable LIBXSMM_PRAGMA if GNU is pre-C99.
  - Fixed issues applying attribute (weak).

* Other
  - MHD: Introduced element-handler info-struct to replace destination-type; allows for conversion hints, etc.
  - MHD: Updated mapping general LIBXSMM data-types to MHD. Support data-comversion handler during MHD-write.
  - Revised implementation of libxsmm_coprime mainly to have more sensible result in case of minco=0.
  - Improved libxsmm_print_cmdline and made prefix/postfix optional arguments.
  - Print PID as part of "Command" info (verbose termination message).
  - Fixed libxsmm_offset and implemented test.
  - Adjusted definition of libxsmm_coprime2.

* Cleanup
  - Deleted libxsmm_ext.h header-file. Removed LIBXSMM_CONST_VOID_PTR.
  - Documentation: updated settings according to documentation (RTD).
  - Documentation: fixed incorrect include-path.
  - Corrected location of header files (VS project).
  - Updated MkDocs.


hfp:
Rebased on top of #838

alheinecke:
Can you please break this PR into multiple smaller one. 

I have seen some transient CI issues which we didn't have on main. I would like to understand if they are related to changes. so breaking into smaller, non-overlapping pieces.

alheinecke:
I have removed some changes which need more work (as the patches very inconsistent). This should also merge well into internal dev branch. 
@egeor : let me know if you see something, otherwise I will merge is this form  which is the majority of the proposed changes.

alheinecke (approved):

egeor (approved):
LGTM

PR #838, opened 2023-12-15
fixed pointer cast of malloc

alheinecke:


hfp:
LGTM. I think this merges well with #837.

Regarding the latter, it resolves the cast-qual issues rather than sprinkling directives all over the code base suppressing the warning. The aforementioned pragmas are currently GCC-specific making the code base warn even more with compilers which are not GCC-compatible.

PR #846, opened 2024-01-25
Feature bitmask-based sparse x dense GEMM

egeor:
This PR adds bitmask-based sparse x dense GEMM. Also, it adds support in the BCSC SPMM with A^T and C^T.


samples/xgemm/kernel_test_bf16only.sh
alheinecke, line 1: It seems this is only minor fix the script I don't understand why we have (bf16 dense only)

alheinecke (approved):
LGTM.

I don't know why we have a bf16 only test script. Do you know? If we don't know we can aso get rid of it.

samples/xgemm/kernel_test_bf16only.sh
egeor: Yeah I had the same question when I looked into this file... It seems to me now it is used only for SRF CI? Can it be that we want to check only bf16 gemms since wee ran through emu (otherwise the codepaths are pretty much the same as avx2)? Similarly there is an f16 only script for GNR.

alheinecke: yeah... right. Forgot about this.

alheinecke: yeah... right. Forgot about this.

PR #848, opened 2024-01-28
Implement correct Intel AMX ABI such that libxsmm can savely interoperate with other SW using Intel AMX

alheinecke:
@egeor : 
- please comment primary on the usage of op.quaternary, if felt this API is OK, It can be seen best in the GEMM driver as now the same gemm_param struct is passed to all 3 kernels (tile config, compute, tile release).
- We could also change to API to two functions: a) tileconfig + new struct and runtime parameter for select to set or to restore b) GEMM kernel without tileconfig.

### DIFF
--- a/samples/spmm/.gitignore
+++ b/samples/spmm/.gitignore
@@ -1,0 +1,3 @@
+spmm_kernel
+kernel_test/.generate_spmm_test_scripts
+kernel_test/*.slurm
--- a/samples/spmm/spmm_kernel.c
+++ b/samples/spmm/spmm_kernel.c
@@ -395,6 +395,7 @@ double jit_matmul( const spmm_def*    i_spmm_def,
   int l_cfg_flags = 0;
   int l_rls_flags = 0;
   unsigned long long N = i_spmm_def->n/i_spmm_def->bn;
+  unsigned char l_tileconfig_save[64] = { 0 };
 
   if (0 == i_spmm_def) {
     fprintf(stderr, "JIT: unsupported descriptor arguments or data type!\n");
@@ -444,18 +445,18 @@ double jit_matmul( const spmm_def*    i_spmm_def,
     exit(-1);
   }
 
-  /* run external tileconfig */
-  if (i_spmm_def->tc_config) {
-    cfg_tr.gemm( NULL );
-  }
-
   /* run correctness */
+  gemm_param.op.quaternary = (void*)l_tileconfig_save;
   gemm_param.a.primary = (void*)i_a;
   gemm_param.b.primary = (void*)i_b;
   gemm_param.b.secondary = (void*)i_colptr;
   gemm_param.b.tertiary  = (void*)i_rowidx;
   gemm_param.b.quaternary = (void*)&N;
   gemm_param.c.primary = (void*)o_c;
+  /* run external tileconfig */
+  if (i_spmm_def->tc_config) {
+    cfg_tr.gemm( &gemm_param );
+  }
   l_test_jit.gemm( &gemm_param );
 
   /* run performance */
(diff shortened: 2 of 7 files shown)


alheinecke:
@ska278 @ddkalamk : FYI and have a look as well

ddkalamk:
So here is my understanding. We need to allocate 64 byte space per thread to save tile config and pass pointer to it in gemm_param structure (gemm_param.op.quaternary). Also, in tile_config and tile_release calls, now we need to pass pointer to gemm_param structure so that it can save and restore correct tile config. Previously, tile_release was a null op.

@alheinecke Is my understanding correct? If this is correct, why not allocate 64 bytes in gemm_param structure itself, instead of allocating and setting pointer separately?

Dhiraj

alheinecke:
@ddkalamk : good suggestion
a) the struct is currently 64bytes: https://github.com/libxsmm/libxsmm/blob/main/include/libxsmm_typedefs.h#L704 so we would double it and every GEMM on any arch will pay the price.
b) we would add it in the end so it's more error-prone in the code gen as we have two GEMM parameter structs (one without fusion and one with fusion).

However, what I want to understand is the difference between the current API proposal and adding a new JITer (just frontend) and having a new struct which is essential just 64byte. But then we would have an additional variable as well. I feel this would be easier for a compiler as this is side-effect free. The current proposal is not fully side-effect free ( I feel).

egeor (approved):
LGTM

alheinecke:
@egeor , @ddkalamk , @ska278 : please have a look at latest update to the PR. As discussed offline, I have created a "new" JITer for tileconfig kernels and the tileconfig size is now opaque.

egeor:
> @ddkalamk : good suggestion a) the struct is currently 64bytes: https://github.com/libxsmm/libxsmm/blob/main/include/libxsmm_typedefs.h#L704 so we would double it and every GEMM on any arch will pay the price. b) we would add it in the end so it's more error-prone in the code gen as we have two GEMM parameter structs (one without fusion and one with fusion).

I guess for (a) this shouldn't be really an issue -- I don't think realistically one would ever allocate huge arrays of gemm_params s.t. the doubling in size would be a problem.

W.r.t. (b), since we touch this new ptr minimally in the code gen (as this PR shows) this shouldn't be really an issue ,  just use the proper offset in the struct pointer (depending on fusion/no fusion just adjust this offset in one line) (unless of course we do more changes to this gemm param struct in the future, but this is not likely to happen?)

```
 
> However, what I want to understand is the difference between the current API proposal and adding a new JITer (just frontend) and having a new struct which is essential just 64byte. But then we would have an additional variable as well. I feel this would be easier for a compiler as this is side-effect free. The current proposal is not fully side-effect free ( I feel).

We could also change to API to two functions: a) tileconfig + new struct and runtime parameter for select to set or to restore b) GEMM kernel without tileconfig.
```

Yeah, for sure this would make the whole thing cleaner as far as the tileconfig business is concerned. 

However, it will be less user friendly, right? I.e. the user that writes platform agnostic code has always to call this new API for set/restore... (even though in other archs it would be a nop, the user still has to write these calls to be truly platform agnostic) 

So, just to be clear, in this case the kernel (b) does not so anything pertaining to tileconfig set/restore, from the outside point of view. If we reconfigure internally in the kernel, then our codegen has to do something in order not to break this API.

alheinecke:
@egeor that's all taken care of in the current PR (or should be :-)).

egeor:
I was confused by your wording "b) GEMM kernel without tileconfig."
I thought you meant internally this (b) would do *nothing* wrt to tileconfig. But apparently this is not what you meant (i.e. when the user creates just a gemm without *anything* in the l_flags, it does internally the store/restore which is the right thing.) I think we are good.

alheinecke:
> BTW, @alheinecke if I pass nullptr to tilecfg kernel, would it fallback to current behavior? i.e. just ignore save restore part? That would be useful as I may want to save/restore only for master thread and ignore for omp worker threads.

No, this is not correct ABI. We don't know what AMX tileconfig was before entering this code. So you still would need to save and restore tileconfig. The old mechanism used tileconfig + tilerelease, which is no longer the correct ABI, so we cannot make use of this mechanism.

PR #852, opened 2024-02-01
[CI] Remove libxsmm-dnn buildkite scripts

rengolin:


alheinecke (approved):

rengolin:
For some reason the tests did not run. I'm not sure why but I'm assuming a glitch in Buildkite. I'll force merge this and if I break the CI on `main`, we can revert and investigate.

PR #858, opened 2024-02-10
Fix SRF CI

alheinecke:
Since migrating to new CI HW (AVX512 based) we experienced transient SRF test failures. These were root caused to a makefile bug not passing the originally set vect arch flags and then reguessing. This patch disables the reguessing as vect arch flags are irrelevant for libxsmm as it is vector code JIT code generator.


rengolin (approved):
Why not just remove the whole thing?

alheinecke:
> Why not just remove the whole thing?

The fix is like advil, it numbs the pain, I believe the bug is still there but we are not triggering it anymore, in case we bother, this code might be helpfuil

PR #871, opened 2024-03-13
Feature mxfp4 bf16 amx gemms

egeor:
This PR introduces mixed precision MXFP4 - BF16 AMX (BR)-GEMM


include/libxsmm_typedefs.h
```diff
@@ -483,6 +483,7 @@ typedef enum libxsmm_gemm_flags {
   LIBXSMM_GEMM_FLAG_INTERPRETE_A_AS_INT4_VNNI2 = 2097152,
   LIBXSMM_GEMM_FLAG_INTERPRETE_A_AS_INT4_VNNI8_INTLV = 4194304,
   LIBXSMM_GEMM_FLAG_DECOMPRESS_A_VIA_BITMASK = 8388608,
+  LIBXSMM_GEMM_FLAG_INTERPRETE_A_AS_MXFP4_VNNI2 = 16777216,
```
alheinecke, line 486: We are running against the wall with flags... I assume the "A_AS" could be done as datatype or are we are running out of bits there as well?

alheinecke (commented):
I'm not happy with the flags as we are using now many flags... any better idea, how we could encode?

include/libxsmm_typedefs.h
egeor: Yeah, using the datatype as intended is the clean solution -- there is no debate... I recommend for this PR to leave this as is, and open an issue to address all these int4/mxfp "A_AS"... After all I want eventually to support all 4 variants of MXFP4 (this is merely a different LUT) and clearly adding extra gemm flags is not scalable... IMHO I think this should be addressed in a different issue/PR (i.e. introduce sub-8 bit dtype)...

alheinecke: Agreed if there is more coming. Opened #872.

alheinecke (approved):
LGTM in principle.

This PR spawned extra future work: #872

PR #881, opened 2024-05-14
Feature mxfp4 bf16 avx2 gemms

egeor:
This PR enables mxfp4-bf16 (BR)gemms for non-AMX as well (for now fallback to avx2) and also generalizes these kernels for any M (before there was m % 8 ==0 requirement)

### DIFF
--- a/src/generator_gemm.c
+++ b/src/generator_gemm.c
@@ -246,9 +246,6 @@ void libxsmm_generator_gemm_kernel( libxsmm_generated_code*        io_generated_
         } else if (l_xgemm_desc_mod.k % 32 != 0) {
           LIBXSMM_HANDLE_ERROR( io_generated_code, LIBXSMM_ERR_ARCH_PREC );
           return;
-        } else if (l_xgemm_desc_mod.m % 8 != 0) {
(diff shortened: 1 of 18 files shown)


src/generator_gemm.c
```diff
@@ -38,6 +38,12 @@ void libxsmm_generator_gemm_kernel( libxsmm_generated_code*        io_generated_
     }
   }
 
+  if ( l_is_Amxfp4_Bbf16_gemm > 0 ) {
+    if (io_generated_code->arch >= LIBXSMM_X86_AVX2 && io_generated_code->arch < LIBXSMM_X86_AVX512_SPR) {
```
alheinecke, line 42: Shouldn't we make the upper limit for now LIBXSMM_X86_ALLFEAT ?

alheinecke (commented):
Minor Nit

src/generator_gemm.c
egeor: This condition basically says if < SPR then enforce AV2 codegen for this kernel -- for AMX we support the kernel with AMX + AVX512 SW pipelined etc.

alheinecke: resolved

alheinecke (approved):

PR #885, opened 2024-05-30
Updated and fixed VS projects.

hfp:


hfp:
A real review may be skipped. Though, this is not actually tested by CI. Further, this change assumes some changes not yet sent as PR.

alheinecke:
if  we need more changes please incl. in the PR.

hfp:
> if we need more changes please incl. in the PR.

It can be merged now "breaks nothing" (or kept until other/entangled changes are PRd).

alheinecke:
closed as it is supersede by #928 which incl. fixes for all additions since May.

PR #887, opened 2024-06-10
Feature int4 gemms scf zpt m x k

egeor:
This PR enables for i4i8 BRGEMMs zero points per MxK block.


include/libxsmm_typedefs.h
```diff
@@ -491,6 +491,8 @@ typedef enum libxsmm_gemm_flags {
   LIBXSMM_GEMM_FLAG_INTERPRETE_A_AS_INT4_VNNI8_INTLV = 4194304,
   LIBXSMM_GEMM_FLAG_DECOMPRESS_A_VIA_BITMASK = 8388608,
   LIBXSMM_GEMM_FLAG_INTERPRETE_A_AS_MXFP4_VNNI2 = 16777216,
+  LIBXSMM_GEMM_FLAG_USE_MxK_ZPT = 33554432,
+  LIBXSMM_GEMM_FLAG_USE_MxK_SCF = 67108864,
```
alheinecke, line 495: For my information: how many bits do we have left?

src/generator_gemm_common.c
```diff
@@ -1366,7 +1366,7 @@ void libxsmm_generator_gemm_setup_stack_frame_fill_ext_gemm_stack_vars( libxsmm_
     libxsmm_generator_gemm_setval_stack_var( io_generated_code, i_micro_kernel_config, LIBXSMM_GEMM_STACK_VAR_B_OFFS_BRGEMM_PTR, temp_reg );
   }
 
-  if (l_is_Ai4_Bi8_gemm > 0 && (i_xgemm_desc->flags & LIBXSMM_GEMM_FLAG_USE_COL_VEC_ZPT) > 0 && (io_generated_code->arch >= LIBXSMM_X86_AVX512_SPR)) {
+  if (l_is_Ai4_Bi8_gemm > 0 && ((i_xgemm_desc->flags & LIBXSMM_GEMM_FLAG_USE_COL_VEC_ZPT) > 0 || (i_xgemm_desc->flags & LIBXSMM_GEMM_FLAG_USE_MxK_ZPT) > 0)) {
```
alheinecke, line 1369: Don't we need at least some > AVX2?

alheinecke (approved):
Please see the two small comments.

include/libxsmm_typedefs.h
egeor: The value on line 495 is 2^26. So basically given the last guard, we should have 4 more bits left.

src/generator_gemm_common.c
egeor: Yeah, I believe even the original arch check was superfluous at this point, we error out based on arch way earlier in the JITer (for good reason).

include/libxsmm_typedefs.h
alheinecke: I think in the same integer some other information is stored... not sure. Need to check the code.

src/generator_gemm_common.c
alheinecke: Ack.

PR #893, opened 2024-08-18
Support for NN, NT, TN, TT GEMMs in FP64, FP32 and BF16

alheinecke:
This PR adds support for NN, NT, TN, TT GEMMs in FP64, FP32 and BF16. In case of BF16 the format of each A / B matrix is flat. The PR aims on faster enabling on AVX512, AVX2-CONVERT and AMX machines such that VNNI packing becomes a performance tweak but not necessary feature.


alheinecke:
@ddkalamk : I have revert the I8 and quant TPP "fixes". These we will fix in the other branch and we will merge the new main into feature_I8_gemm ASAP when this PR is merged

egeor (approved):
LGTM

PR #895, opened 2024-08-31
Feature quant dequant i8 avx2

alheinecke:


egeor:
Should be good @alheinecke, latest commits also include support for truncation with signed saturation.

egeor (approved):

PR #909, opened 2024-10-11
Parallel TPP Test Script Generation

alheinecke:
Significant time of setting up testing of TPP in CI is spent on generating scripts. Especially on VMs/Emulation.


egeor (approved):
LGTM

PR #910, opened 2024-10-16
Adding support for M4 SME gemm

stefan0re:
This PR contains kernel generation for C+=A*B and C+=A*B^T, on an M4 which supports SME.

- Supports just FP32
- Supports leading dimensions not matching M, N, K


src/generator_aarch64_instructions.c
```diff
@@ -2182,3 +2184,305 @@ void libxsmm_aarch64_instruction_cond_jump_to_label( libxsmm_generated_code*
...
+    LIBXSMM_HANDLE_ERROR( io_generated_code, LIBXSMM_ERR_BUFFER_TOO_SMALL );
+    return;
+  }
+
+  unsigned int code_head = io_generated_code->code_size/4;
+  unsigned int* code     = (unsigned int *)io_generated_code->generated_code;
+  /* fix bits */
+  code[code_head] = i_instr;
+
+  if( i_instr == LIBXSMM_AARCH64_INSTR_SME_MOVA_32_BIT_TILE_TO_VECTOR || i_instr == LIBXSMM_AARCH64_INSTR_SME_MOVA_V_TILE_TO_VECTOR ){
```
alheinecke, line 2324: it would be good to avoid if conditions based on instruction type and if we could use some bits which will be overriden (register number etc.) to carry this information.

src/generator_common_aarch64.c
```diff
@@ -4082,3 +4082,1575 @@ void libxsmm_aarch64_instruction_sve_memcpy( libxsmm_generated_code*        io_g
...
+                                                LIBXSMM_AARCH64_INSTR_GP_MOVZ,
+                                                LIBXSMM_AARCH64_GP_REG_W12+i,
+                                                0,
+                                                i );
+  }
+  if( i_m_blocking == 32 && i_n_blocking == 32 ){
+    libxsmm_aarch64_instruction_set_ptrue_as_counter_sve2( io_generated_code,
+                                                           LIBXSMM_AARCH64_INSTR_SVE2_PTRUE_AS_COUNTER,
+                                                           LIBXSMM_AARCH64_SVE_REG_P8 );
+    for( int l = 0; l < 2; l++){
```
alheinecke, line 4104: nit: let's use unsigned int for all loop counter variables, the code is mixing

src/generator_common_aarch64.c
```diff
@@ -4082,3 +4082,1575 @@ void libxsmm_aarch64_instruction_sve_memcpy( libxsmm_generated_code*        io_g
                                                    i_gp_reg_dst, i_gp_reg_dst, l_incremented_size, 0 );
   }
 }
+LIBXSMM_API_INTERN
```
alheinecke, line 4085: General comment to all of the routines (load, store, transpose): I assume you have looked into the possibility of modularization, but it's to special coded such that it doesn't work. It felt when reading the code, that there was a lot of repetition / manual duplication, but I tried at 1 or 2 places and could not easily see how we could modularization to reduce the the diff from 1.5K LoC to something substantially smaller. Would you agree?

src/generator_gemm.c
```diff
@@ -785,6 +788,8 @@ void libxsmm_generator_gemm_kernel( libxsmm_generated_code*        io_generated_
     libxsmm_generator_gemm_aarch64_kernel( io_generated_code, &l_xgemm_desc_mod );
   } else if ( (io_generated_code->arch == LIBXSMM_AARCH64_SVE512) || (io_generated_code->arch == LIBXSMM_AARCH64_A64FX) ) {
     libxsmm_generator_gemm_aarch64_kernel( io_generated_code, &l_xgemm_desc_mod );
+  } else if ( io_generated_code->arch == LIBXSMM_AARCH64_APPL_M4 ) {
+    libxsmm_generator_gemm_aarch64_kernel_sme_het_blocking( io_generated_code, &l_xgemm_desc_mod );
```
alheinecke, line 792: Should this be guarded for F32 only as the PR description says FP32 only?

include/libxsmm_cpuid.h
```diff
@@ -47,6 +47,7 @@
 #define LIBXSMM_AARCH64_NEOV1         2302 /* Neoverse V1, Graviton 3 */
 #define LIBXSMM_AARCH64_SVE512        2401 /* SVE 512 */
 #define LIBXSMM_AARCH64_A64FX         2402 /* A64FX */
+#define LIBXSMM_AARCH64_APPL_M4       2501 /* Apple M4 SME without SVE*/
```
alheinecke, line 50: I was surprised that so little other conditions changed in ARCH selection. I assume FP64, BF16 still works on M4 using ASIMD?

src/libxsmm_cpuid_arm.c
```diff
@@ -144,7 +144,8 @@ LIBXSMM_API int libxsmm_cpuid_arm(libxsmm_cpuid_info* info)
 # else
     void (*const handler)(int) = signal(SIGILL, internal_cpuid_arm_sigill);
 #   if defined(__APPLE__) && defined(__arm64__)
-    result = LIBXSMM_AARCH64_APPL_M1;
+    result = LIBXSMM_AARCH64_APPL_M4;
```
alheinecke, line 147: Can we add an env variable, just overriding M1 to M4 doesn't seem to be good as there are many M1,M2 out there.

alheinecke (commented):
Let some initial questions.

src/generator_aarch64_instructions.c
```diff
@@ -936,11 +936,13 @@ void libxsmm_aarch64_instruction_sve_move( libxsmm_generated_code*
                                            const int                              i_offset,
                                            const unsigned int                     i_vec_reg,
                                            const unsigned int                     i_pred_reg ) {
+  /*
```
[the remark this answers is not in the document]
stefan0re: Yes, it was there because of an early state of the code, i will fix it

src/generator_common_aarch64.c
stefan0re: Yes I agree, will take care of it.

alheinecke:
Great, please let me know when you are ready I think you also need to merge main back into your branch. Then we can create a branch for M4 and run CI.

stefan0re:
I have merged the main branch into mine, so it should be possible to create a branch and run CI. I wasn't able to reduce the LoC for the load, store, and transpose functions.

alheinecke:
I changed to based to "feature_m4_base" which is a copy of main, however, there are still conflicts hence I couldn't merge.

src/generator_common_aarch64.c
alheinecke: Is this taken care of as well?

PR #911, opened 2024-10-17
Feature spr gemm flat a sw pipelined

egeor:
On AMX we add support for flat A with online vnni convert via SW pipeline

### DIFF
--- a/src/generator_gemm_common.c
+++ b/src/generator_gemm_common.c
@@ -472,7 +472,7 @@ LIBXSMM_API_INTERN void libxsmm_generator_gemm_apply_opA_opB( libxsmm_generated_
   if ( ( i_xgemm_desc_orig->flags & LIBXSMM_GEMM_FLAG_TRANS_A) && (i_xgemm_desc_orig->m != 0) && (i_xgemm_desc_orig->k != 0) ) {
     /* if A needs to be transposed, use scratch in stack */
     libxsmm_generator_gemm_setup_A_vnni_or_trans_B_vnni_or_trans_tensor_to_stack( io_generated_code, io_loop_label_tracker, i_gp_reg_mapping, i_micro_kernel_config, i_xgemm_desc, i_xgemm_desc_orig, (libxsmm_datatype) LIBXSMM_GEMM_GETENUM_AB_COMMON_PREC( i_xgemm_desc_orig->datatype ), 0);
-  } else if ( (i_micro_kernel_config->avnni_gemm_stack_alloc_tensors !=0) || (i_micro_kernel_config->avnni_btrans_gemm_stack_alloc_tensors !=0) ) {
+  } else if ( (i_micro_kernel_config->avnni_gemm_stack_alloc_tensors !=0 && (io_generated_code->arch < LIBXSMM_X86_AVX512_SPR)) || (i_micro_kernel_config->avnni_btrans_gemm_stack_alloc_tensors !=0) ) {
     /* if B is in vnni2T, use scratch in stack */
     libxsmm_generator_gemm_setup_A_vnni_or_trans_B_vnni_or_trans_tensor_to_stack( io_generated_code, io_loop_label_tracker, i_gp_reg_mapping, i_micro_kernel_config, i_xgemm_desc, i_xgemm_desc_orig, (libxsmm_datatype) LIBXSMM_GEMM_GETENUM_AB_COMMON_PREC( i_xgemm_desc_orig->datatype ), 0);
   } else if ( ((i_xgemm_desc_orig->flags & LIBXSMM_GEMM_FLAG_TRANS_B) > 0) && ((i_xgemm_desc_orig->flags & LIBXSMM_GEMM_FLAG_VNNI_B) > 0) ) {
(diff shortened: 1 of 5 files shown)


src/generator_gemm_common.c
```diff
@@ -472,7 +472,7 @@ LIBXSMM_API_INTERN void libxsmm_generator_gemm_apply_opA_opB( libxsmm_generated_
   if ( ( i_xgemm_desc_orig->flags & LIBXSMM_GEMM_FLAG_TRANS_A) && (i_xgemm_desc_orig->m != 0) && (i_xgemm_desc_orig->k != 0) ) {
     /* if A needs to be transposed, use scratch in stack */
     libxsmm_generator_gemm_setup_A_vnni_or_trans_B_vnni_or_trans_tensor_to_stack( io_generated_code, io_loop_label_tracker, i_gp_reg_mapping, i_micro_kernel_config, i_xgemm_desc, i_xgemm_desc_orig, (libxsmm_datatype) LIBXSMM_GEMM_GETENUM_AB_COMMON_PREC( i_xgemm_desc_orig->datatype ), 0);
-  } else if ( (i_micro_kernel_config->avnni_gemm_stack_alloc_tensors !=0) || (i_micro_kernel_config->avnni_btrans_gemm_stack_alloc_tensors !=0) ) {
+  } else if ( (i_micro_kernel_config->avnni_gemm_stack_alloc_tensors !=0 && (io_generated_code->arch < LIBXSMM_X86_AVX512_SPR)) || (i_micro_kernel_config->avnni_btrans_gemm_stack_alloc_tensors !=0) ) {
```
alheinecke, line 475: I'm not sure if this logic is the best. Wouldn't it be better to change it here?
https://github.com/libxsmm/libxsmm/blob/main/src/generator_gemm_amx.c#L3068

Or is this as we still need the allocation, but we don't need to norm->vnni transform anymore... but then still just comparing for smaller SPR feels to general.

src/generator_gemm_common.c
```diff
@@ -1536,7 +1536,7 @@ void libxsmm_generator_gemm_setup_stack_frame_allocate_scratch( libxsmm_generate
     libxsmm_generator_gemm_setval_stack_var( io_generated_code, i_micro_kernel_config, LIBXSMM_GEMM_STACK_VAR_GEMM_SCRATCH_PTR, LIBXSMM_X86_GP_REG_RSP );
   }
 
-  if ((io_generated_code->arch >= LIBXSMM_X86_AVX512_SPR) && (l_is_Ai4_Bi8_gemm > 0 || l_is_Abf8_Bbf16_gemm > 0 || l_is_Abf8_Bf16_gemm > 0 || l_is_Ahf8_Bbf16_gemm > 0 || l_is_Amxfp4_Bbf16_gemm > 0)) {
+  if ((io_generated_code->arch >= LIBXSMM_X86_AVX512_SPR) && (l_is_Ai4_Bi8_gemm > 0 || l_is_Abf8_Bbf16_gemm > 0 || l_is_Abf8_Bf16_gemm > 0 || l_is_Ahf8_Bbf16_gemm > 0 || l_is_Amxfp4_Bbf16_gemm > 0 || i_micro_kernel_config->avnni_gemm_stack_alloc_tensors > 0)) {
```
alheinecke, line 1539: why can't the stack allocation from the if not being used?

src/generator_gemm_common.c
```diff
@@ -1546,7 +1546,7 @@ void libxsmm_generator_gemm_setup_stack_frame_allocate_scratch( libxsmm_generate
   }
 
   if ((i_micro_kernel_config->bf8_gemm_via_stack_alloc_tensors > 0) || (i_micro_kernel_config->hf8_gemm_via_stack_alloc_tensors > 0) ||
-      (i_micro_kernel_config->atrans_gemm_stack_alloc_tensors > 0 ) || (i_micro_kernel_config->avnni_gemm_stack_alloc_tensors > 0) ||
+      (i_micro_kernel_config->atrans_gemm_stack_alloc_tensors > 0 ) || (i_micro_kernel_config->avnni_gemm_stack_alloc_tensors > 0 && (io_generated_code->arch < LIBXSMM_X86_AVX512_SPR)) ||
```
alheinecke, line 1549: see above first comment.

src/generator_gemm_common.c
alheinecke: I think we are now allocating twice?

alheinecke (commented):
The AMX kernel itself looks fine, I think we should improve the condition when calling the stack conversion, just checking for smaller SPR feels to general, eg when we add BF8 / FP8 flat it will break and we will hunt this condition. Perhaps an additional field in the trashcan?

src/generator_gemm_common.c
egeor: I guess we could add atop BF16 and F16 conditions

src/generator_gemm_common.c
egeor: Nope, the second "big allocation" happens only for < spr

src/generator_gemm_common.c
egeor: These would have different sizes, thus I follow the logic for all "A SW pipeline" where the space is MxK

src/generator_gemm_common.c
alheinecke: I think it's better to add a flag to micro_kernel_config like "swpipeline_vnni_transform", then we also are future proof for optimizations for BFMMLA etc. where SW Pipeline would make sense as well.

src/generator_gemm_common.c
alheinecke: Ah I see there is an allocation above, but it can be guarded with the same "swpipeline_vnni_transform" flag, I assume.

src/generator_gemm_common.c
alheinecke: and here we can use the same flag again, correct?

alheinecke (approved):
LGTM

PR #920, opened 2024-11-12
Added fallback to NEON for M4 eltwise kernel

stefan0re:
Now the eltwise kernels run with Neon code until they are implemented with SME.
Previously, no eltwise kernel could be called for M4 because it was not detected correctly.


stefan0re:
The list with the allowed kernels for Neov2 has been adapted so that mmla and bf16 flatgemm are also tested

alheinecke (approved):
LGTM

alheinecke:
@stefan0re : one nit I just saw, can you please update the neoverseV2 allow list and sort this alphabetically?

alheinecke (approved):

PR #926, opened 2024-12-16
Feature reference jit kernels

egeor:
This PR adds reference kernels for gemm/eltwise/equation TPPs; in case the performance-oriented/platform specific JIT path fails, the fallback reference kernel is used.


src/generator_aarch64_reference.c
```diff
@@ -0,0 +1,213 @@
...
+  }
+  libxsmm_aarch64_instruction_alu_compute_imm12( io_generated_code, LIBXSMM_AARCH64_INSTR_GP_ADD_I, LIBXSMM_AARCH64_GP_REG_XSP, l_temp_reg2, 0, 0 );
+  libxsmm_aarch64_instruction_alu_compute_imm64( io_generated_code, LIBXSMM_AARCH64_INSTR_GP_META_SUB,
+                                                 l_temp_reg2, l_temp_reg, l_temp_reg2,
+                                                 (long long) l_padded_desc_size );
+  libxsmm_aarch64_instruction_alu_compute_imm12( io_generated_code, LIBXSMM_AARCH64_INSTR_GP_ADD_I, l_temp_reg2, LIBXSMM_AARCH64_GP_REG_XSP, 0, 0 );
+  libxsmm_aarch64_instruction_alu_compute_imm12( io_generated_code, LIBXSMM_AARCH64_INSTR_GP_ADD_I, LIBXSMM_AARCH64_GP_REG_XSP, l_temp_reg, 0, 0 );
+  /* We set the address of the function  */
+  if (i_is_gemm_or_eltwise == 0) {
+    libxsmm_aarch64_instruction_alu_set_imm64( io_generated_code, l_temp_reg2, (unsigned long long) libxsmm_reference_gemm  );
```
alheinecke, line 81: Can we switch to using the same union as in x86 case?
Is the type of the libxsmm_aarch64_instruction_alu_set_imm64 unsigned or signed?

src/libxsmm_main.c
```diff
@@ -2112,6 +2158,12 @@ LIBXSMM_API_INTERN int libxsmm_build(const libxsmm_build_request* request, unsig
   char jit_buffer[LIBXSMM_CODE_MAXSIZE] = { 0 }, jit_name[384] = { 0 }, suffix_name[16] = { 0 };
   libxsmm_generated_code generated_code /*= { 0 }*/;
   libxsmm_kernel_xinfo extra /*= { 0 }*/;
+  const char *const env_reference_disable_gemm_fallback = getenv("LIBXSMM_DISABLE_GEMM_REFERENCE_FALLBACK");
```
alheinecke, line 2161: Can we add these env variable readings to all the other env variables and query through a function call?

src/libxsmm_utils.c
```diff
@@ -17,6 +17,13 @@
 # define LIBXSMM_RNG_DRAND48
 #endif
 
+LIBXSMM_API int libxsmm_return_success_code(unsigned int i_used_reference_kernel) {
```
alheinecke, line 20: see above... I know why it's here... I'm not sure if we should provide a value the "reference" in the main library as this is more a testing trick we are using to get the output correct, but I feel that doesn't need to ship with the main library. I know that would mean some duplication, but I feel it;s clear to add this to "samples/eltwise/eltwise_common.h" and the "samples/xgemm/gemm_kernel.c" and we anyways need to make those two reusing some of the other infra as well.

alheinecke (commented):
Very nice! 

I have some minor nits.

egeor:
> Very nice!
> 
> I have some minor nits.

Should be good to go now.

alheinecke (approved):

PR #929, opened 2024-12-29
Increasing #elements/#pointers in matrix arg struct to 6

alheinecke:
In order to support dynamic LDs we need to get more arguments into the kernel. As we are at the limit of elements this PR increases the number arguments from four to six, but no support for dynamic LDs is added. The goal of the PR is to enable apps using libxsmm to adopt to the new API (which should be zero code change) and to test with the new version of libxsmm

### DIFF
--- a/include/libxsmm.h
+++ b/include/libxsmm.h
@@ -425,9 +425,9 @@ template<int PREFETCH> inline/*superfluous*/ void libxsmm_mmfunction_prefetch(
   LIBXSMM_ASSERT(LIBXSMM_GEMM_PREFETCH_NONE != PREFETCH);
   if (0/*EXIT_SUCCESS*/ == libxsmm_get_mmkernel_info(xmm, &info) && LIBXSMM_GEMM_PREFETCH_NONE != info.prefetch) {
     const size_t itypesize = LIBXSMM_TYPESIZE(info.iprecision), otypesize = LIBXSMM_TYPESIZE(info.oprecision);
-    args.a.quaternary = static_cast<char*>(args.a.primary) + itypesize * info.m * info.k;
-    args.b.quaternary = static_cast<char*>(args.b.primary) + itypesize * info.k * info.n;
-    args.c.quaternary = static_cast<char*>(args.c.primary) + otypesize * info.m * info.n;
+    args.a.senary = static_cast<char*>(args.a.primary) + itypesize * info.m * info.k;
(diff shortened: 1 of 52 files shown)


rengolin (approved):

egeor (approved):
Looks good to me. The only general comment is that in this PR the "magic constants" are "just" adjusted properly (vs adding some named constants throughout). For now we should be good as discussed earlier (as most likely we will not touch this again) but maybe as an enhancement we can add an issue to keep track of this (and link to this PR such that the touch points are obvious)

alheinecke:
Yes, I'm right now working on a change set to make them all sizeof(libxsmm_matrix_arg) and some calculations, will hopefully land tonight PT in the PR and then I will merge. 

The only thing it won't fix is the equation scratch stuff as this still works with magic pointers.

rengolin:
Yeah, not super happy about the constants either, but I guess that's more than just a little refactoring that will need its own PR later.

ddkalamk:
My understanding is application code change is required only if GEMM PREFETCH flag is used. Since I am not using any prefetch flags, I don't need any change in extension code to use new APIs.

@alheinecke is my understanding correct?

alheinecke:
> My understanding is application code change is required only if GEMM PREFETCH flag is used. Since I am not using any prefetch flags, I don't need any change in extension code to use new APIs.

Yepp, that's correct. And of course since the recommendation is to memset to zero all the parameter structs before handing them to the JIT'ed kernels the zero-ing is automatically adjusted as one uses `sizeof(struct)`.

However, since this PR is touching the guts of LIBXSMM, I recommend updating to latest main after merge ASAP and testing if everything is still working as expected. 

@egeor , @rengolin : I have just pushed more changes which eliminate ~90% of the magic number usages and makes future changes to the function parameter structs very easy. 

@egeor : feel free to adopt the define scheme to remove the other magic numbers as well. As we all agree this is not urgent and after merge I will create an issue with a link to the one define I have already created within this PR.

alheinecke (commented):
just reviewed again in the GUI, looks good

ddkalamk:
> However, since this PR is touching the guts of LIBXSMM, I recommend updating to latest main after merge ASAP and testing if everything is still working as expected.

@alheinecke Just tested llm example from tpp pytorch extension with latest libxsmm main, everything looks good!

PR #931, opened 2025-01-13
[Github Actions] Add test for AWS builder

rengolin:
This is a temporary fix to move to Github Actions. It does not start/stop the instance automatically, this is left for later once we get the IAM/secret keys right, but it works as a temporary stop-gap for now.

I have added triggers for `workflow_dispatch` (ie. press the button on Github's web interface) and `push` (which also counts PRs). This means for now we still need to activate the instance whenever it runs, so we could keep only on `worflow_dispatch` and not `push`?

I don't mind either way.

### DIFF
--- a/.github/workflows/libxsmm-arm-graviton3.yml
+++ b/.github/workflows/libxsmm-arm-graviton3.yml
@@ -9,8 +9,26 @@ env:
   SYM: 2
 
 jobs:
+  ec2-start:
+    runs-on: ubuntu-latest
+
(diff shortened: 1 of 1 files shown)


alheinecke:
Is there any chance we can get automatic start going by early next week (01/20).

rengolin:
> Is there any chance we can get automatic start going by early next week (01/20).

Done.

alheinecke (approved):
LGTM

PR #932, opened 2025-01-20
Initial Support for Static Code Generation via Binary Export

alheinecke:
This PR lays the ground work for retiring the old static generator (inline assembly, x86 ore SPR  only) and using the binary generator instead. Therefore following tools are added through this PR:
- a helper application that can drive the libxsmm code generators to dump the binary to disk
- a python script to created static libraries and headers via the aforementioned tool
- a simple tester for the added infrastructure


adam-smnk:
> Alternatively, we could use the Python wrapper to generate a static library with the most common variations for particular models and just link them automatically (as we do today) at run time.

I think that would be easiest and most desirable PoC toward product requirements. For that, the current python script seems enough.
Compiler shouldn't need to know anything about available options. (Fat) Library wrapper can hide these details and at best provide a util like `is_config_supported` to allow for fallback to some default lowering.

rengolin:
> I think that would be easiest and most desirable PoC toward product requirements. For that, the current python script seems enough.

For a PoC, I agree.

But as we increase the usage, especially with mixed precision, this can escalate quickly. If we are to propose such a solution, we at least must have a prototype on how such a dynamic library generation would work.

scripts/libxsmm_staticlib.py
```diff
@@ -0,0 +1,465 @@
...
+        ar_cmd = self.getArCmd(archivefile, objfiles)
+        self.logger.info(F"ar cmd: {ar_cmd}")
+        res = self.runner.run(ar_cmd)
+        self.logger.info(res.stdout)
+        if (res.stderr != ""):
+             self.logger.error(res.stderr)
+             sys.exit(1)
+
+        headerfile.write("\n#ifdef __cplusplus\n")
+        headerfile.write("extern }\n")
```
adam-smnk, line 349: nit: extern not needed here; `} // extern "C"` could be nice

adam-smnk:
> The compiler finds dispatch / invoke pairs for the kernels, just like we do today (even as function calls) and registers them all as usual.

At this point, we can collapse these into a single call.

> But before we JIT the rest of the code, we collect all the functions and dispatch the C generator to compile them all and emit a shared object that we `dlopen` just before invoking the JITer?

Let's assume we can't or won't tightly integrate the code gen into an external runner.
We'd probably need to provide our own caching mechanism as a part of compiler's xsmm runtime utils. Basically what dispatch already does today but with persistent dump to disk and lookup on init.

rengolin:
> > The compiler finds dispatch / invoke pairs for the kernels, just like we do today (even as function calls) and registers them all as usual.
> 
> At this point, we can collapse these into a single call.

Are you proposing all kernel arguments to become parameters and the single function becomes a dispatcher, just like `statictest.c`? This would simplify the API and the compiler test.

The negative side is that we're adding a long branch and a few comparisons to every kernel call, which happens into every inner loop iteration, and without some powerful recursive inlining in LTO, these will not be removed. 

> > But before we JIT the rest of the code, we collect all the functions and dispatch the C generator to compile them all and emit a shared object that we `dlopen` just before invoking the JITer?
> 
> Let's assume we can't or won't tightly integrate the code gen into an external runner. We'd probably need to provide our own caching mechanism as a part of compiler's xsmm runtime utils. Basically what dispatch already does today but with persistent dump to disk and lookup on init.

One goal of this proposal would be to design an API that allows us to do that. Whether we want to push this upstream or give up will depend on how generic AND successful it is. For that, we'll need to look into other types of micro-kernel libraries.

adam-smnk:
> Are you proposing all kernel arguments to become parameters and the single function becomes a dispatcher, just like statictest.c? This would simplify the API and the compiler test.
The negative side is that we're adding a long branch and a few comparisons to every kernel call, which happens into every inner loop iteration, and without some powerful recursive inlining in LTO, these will not be removed.

Actually, there are two cases.

This is assuming the dispatch has dynamic arguments. From ease of integration perspective, it could be slightly more convenient and gets you 90% there. On the other hand, injecting dispatch like we do currently is already pretty trivial so maybe it's not really that beneficial.

But in cases where dispatch arguments are static (or directly matmul op has all static info) like in Triton, we could generate the code at compile time and append it to the kernel. We risk some bloat due to codegen duplication but it avoids long branches and/or LTO could more easily clean it up.

alheinecke:
The PR is just about generating static library with the code gen infra. The allows to create another project to build a dynamic shape library based on a finite set (which can still very large) of static shape & dynamic stride kernels. Not sure why we discuss how to use this functionality here, as it IMHO is not part of libxsmm.

scripts/libxsmm_staticlib.py
alheinecke: thanks and done.

rengolin (approved):
Yeah, ignore our tangent, we just went wild. LGTM, thanks!

Issue #933, opened 2025-01-24
Build failure with GCC 15


loveshack:
GCC 15 changes the default C language standard to C23. Building with, say, -std=gnu17 will be needed to avoid these build failures seen with a pre-release in Fedora:

```
./src/libxsmm_ext.c: In function ‘dgemm_batch_’:
./src/libxsmm_ext.c:182:3: error: too many arguments to function ‘internal_noblas_error((const char *)"dgemm_batch")’; expected 0, have 15
  182 |   internal_noblas_error("dgemm_batch")(transa_array, transb_array, m_array, n_array, k_array,
      |   ^~~~~~~~~~~~~~~~~~~~~                ~~~~~~~~~~~~
./src/libxsmm_ext.c: In function ‘sgemm_batch_’:
./src/libxsmm_ext.c:196:3: error: too many arguments to function ‘internal_noblas_error((const char *)"sgemm_batch")’; expected 0, have 15
  196 |   internal_noblas_error("sgemm_batch")(transa_array, transb_array, m_array, n_array, k_array,
      |   ^~~~~~~~~~~~~~~~~~~~~                ~~~~~~~~~~~~
./src/libxsmm_ext.c: In function ‘dgemm_’:
./src/libxsmm_ext.c:209:3: error: too many arguments to function ‘internal_noblas_error((const char *)"dgemm")’; expected 0, have 13
  209 |   internal_noblas_error("dgemm")(transa, transb, m, n, k, alpha, a, lda, b, ldb, beta, c, ldc);
      |   ^~~~~~~~~~~~~~~~~~~~~          ~~~~~~
./src/libxsmm_ext.c: In function ‘sgemm_’:
./src/libxsmm_ext.c:222:3: error: too many arguments to function ‘internal_noblas_error((const char *)"sgemm")’; expected 0, have 13
  222 |   internal_noblas_error("sgemm")(transa, transb, m, n, k, alpha, a, lda, b, ldb, beta, c, ldc);
      |   ^~~~~~~~~~~~~~~~~~~~~          ~~~~~~
./src/libxsmm_ext.c: In function ‘dgemv_’:
./src/libxsmm_ext.c:233:3: error: too many arguments to function ‘internal_noblas_error((const char *)"dgemv")’; expected 0, have 11
  233 |   internal_noblas_error("dgemv")(trans, m, n, alpha, a, lda, x, incx, beta, y, incy);
[...]

PR #934, opened 2025-01-25
make -std=c99 default for GCC

alheinecke:
Addressing: #933


rengolin (approved):
I guess in 2035 we'll change it to `c11`. :rofl:

PR #937, opened 2025-01-29
Clean-Up AVX512 GEMM Code generator

alheinecke:
In order to prepare a cleaner dynld integration, this PR refactors / minimizes code paths in the AVX512 GEMM code generator:

- Refactor AVX512 Fused Broadcast FMA generator path: avoid going for precision emulation cases into this routine
- removed emulation of BF8 with inline instructions, BF8 is aligned with HF8 and both BF8 and HF8 are now both implemented using stack scratch memory.


egeor (approved):
Looks good to me!

PR #938, opened 2025-01-30
removing b pointers fused address calculation for very small K

alheinecke:
For very small K (<24) the x86 code generators avoid on GPR add per M/N block. It was not measurable that this give any performance benefit, but it complicates code generation and increases code to maintain. This PR removes this optimization.


egeor (approved):
LGTM

PR #944, opened 2025-02-12
Compilation and pedantic fixes.

WillTrojak:
This should fix the compilation and pedantic errors on ppc64le.


WillTrojak:
@alheinecke

PR #945, opened 2025-02-20
Adding support for SME on M4

stefan0re:
- Adding CPU detection
- Most of the kernels run with SVE 512 in Streaming Mode
- Some kernels fall back to NEON due to unsupported instructions in Streaming Mode.


alheinecke:
I'm merging into feature_m4 branch such that we can track CI

PR #947, opened 2025-02-25
Adding support for zarch/s390x

WillTrojak:
Adds support for zarch/s390x for zarch=11 to zarch=14 (ie z13 to the current z16), with some future proofing concepts. Prior to z13, there weren't vector facilities, and zEC12 is out of support now.

This PR adds support for non-transposed FP32 and FP64 gemm kernels, with measured FLOPs about 15% higher than OpenBLAS. The code currently compiles without errors on max pedantic for both z15 and z16. 

Todo is adding reference trampoline functions. 

One issue that I was not able to solve is the syncing. It would be good to have the input of @alheinecke on the the syncing code. Currently, I detect s390x and disable syncing, similar to what seems to be done for Cray.

Through IBM LinuxONE Community Cloud I have secured free z15 access for development and CI. @alheinecke and @rengolin, if you contact me, I can add you to the instance.


WillTrojak:
@Andreas-Krebbel

src/generator_s390x_instructions.c
```diff
@@ -0,0 +1,3489 @@
...
+  o_reg.vr = l_vr;
+  return o_reg;
+}
+
+LIBXSMM_API_INTERN
+libxsmm_s390x_reg_stack* libxsmm_s390x_reg_stack_init( unsigned int            i_n,
+                                                       libxsmm_s390x_reg_type  i_reg_type,
+                                                       unsigned int           *i_resv,
+                                                       unsigned int            i_nresv ) {
+  libxsmm_s390x_reg_util *l_util = malloc(i_n*sizeof(enum libxsmm_s390x_reg_util));
```
FreddieWitherden, line 818: Prefer calloc as it zeros.  Check for NULL returns and error.

src/generator_s390x_instructions.c
```diff
@@ -0,0 +1,3489 @@
...
+                              unsigned int   i_2,
+                              unsigned int   i_3,
+                              unsigned int   i_4,
+                              unsigned char *o_out ) {
+  char l_nbytes;
+
+  unsigned long l_fid = i_instr & LIBXSMM_S390X_FMASK;
+  unsigned long l_instr = i_instr & ~LIBXSMM_S390X_FMASK;
+  switch( l_fid ) {
+    case LIBXSMM_S390X_FORM_RIE_F_FORM: {
```
FreddieWitherden, line 1700: I'd see if you can refactor these kinds of blocks to use a static table of function pointers.  May need to tweak the arguments a bit so they're all compatible.

src/generator_s390x_instructions.c
```diff
@@ -0,0 +1,3489 @@
...
+output[3] = 0xff & (opcode >> 16);
+output[4] = 0xff & (opcode >> 8);
+output[5] = 0xff & (opcode >> 0);
+return 6;
+}
+
+LIBXSMM_API_INTERN
+char libxsmm_s390x_instr_rxy_b_form(unsigned long instr, unsigned char m1,unsigned char x2,unsigned char b2,unsigned int dl2,unsigned char dh2, unsigned char *output) {
+unsigned long opcode = (0xffffffffffff & instr);
+opcode += (unsigned long)(0xf & m1) << 36;
```
FreddieWitherden, line 2798: Can you not just simplify these functions to create a long opcode and then factor out the loop?

alheinecke:
I have changed the base to copy of main such that we can merge and run CI for ARM, RV64 and x86. However, @WillTrojak let's first address the comments.

src/generator_s390x_instructions.c
FreddieWitherden: Actually, it might be better to make all of these little functions macros which will reduce overhead and ensure everything gets inlined.

uweigand:
> Adds support for zarch/s390x for zarch=11 to zarch=14 (ie z13 to the current z16), with some future proofing concepts. Prior to z13, there weren't vector facilities, and zEC12 is out of support now.

Thanks for working on IBM Z support!   If there's any platform-specific questions I can help with, please let me know.
 
> One issue that I was not able to solve is the syncing. It would be good to have the input of @alheinecke on the the syncing code. Currently, I detect s390x and disable syncing, similar to what seems to be done for Cray.

We should fully support the same set of synchronization primitives (__sync_... and __atomic_... builtins) with GCC and LLVM on s390x that most other Linux targets do.   Is there anything that was missing or causing issues?

WillTrojak:
@FreddieWitherden I did some refactoring of the instruction formation. I think it's a lot cleaner now. Just so you know, there are two considerations on Z, instructions can be 2, 4 or 6 bytes long and the system is big-endian. Also, some instructions have an implicit RBX field, that I move into the instruction formation. This cleans up the code, but also makes some of the deferred instructions much easier. I'll make the tool I made to write these sections public in the coming days.

WillTrojak:
@uweigand there is a load of synchronization code here that was the issue: https://github.com/libxsmm/libxsmm/blob/29d490b99efac57e1c1b8d4aa472b0c1d7d15cad/include/libxsmm_sync.h

From what I can understand, this uses atomics for thread safety at generation time. However, before the lines I added permitting unsafe generation on Z, the generation would hang. The code here is pretty dense CPP macros so is quite hard to follow. @alheinecke any help you can offer in terms of what is required in this section, would be great.

alheinecke:
@WillTrojak : what is exactly the issue with locks (as the code works with builtins and was fine on x86, arm, rv64, ppc64le)

FreddieWitherden (approved):

PR #949, opened 2025-03-06
Add Mac M4 support

alheinecke:


alheinecke:
@stefan0re : looks good, I have opened a PR from the branch.

src/generator_common_aarch64.c
```diff
@@ -241,12 +241,13 @@ void libxsmm_generator_gather_scatter_vreg_asimd_aarch64( libxsmm_generated_code
                                                           const unsigned int      i_masked_elems,
                                                           const unsigned int      i_is_gather ) {
   unsigned int i = 0;
+  unsigned int l_vlen = (io_generated_code->arch != LIBXSMM_AARCH64_APPL_M4 ) ? libxsmm_cpuid_vlen(io_generated_code->arch) : 16;
```
egeor, line 244: Can't we hide this decision within the libxsmm_cpuid_vlen call?

egeor (approved):
Looks good to me. Just a minor nit wrt a vlen decision and if we can hide that under libxsmm_cpuid_vlen.

src/generator_common_aarch64.c
stefan0re: The problem is that in this function I need the NEON vector length whereas almost all other functions need the SME vector length.

alheinecke: @stefan0re : I have looked more: https://github.com/libxsmm/libxsmm/blob/feature_m4/src/libxsmm_cpuid_x86.c#L580 seems to be incorrect. Apple M4 should return 4 and not 16 and then this changes is not needed. Why does cpuid_vlen return 16 for M4 and not 4? The machine has only asimd, but no SVE.

stefan0re: M4 has a streaming SVE mode to use SME and this has 512-bit vectors.
Not 100% of the SVE instructions are supported there, which means I have to switch to NEON and use the other vector length.

alheinecke:
I assume when we run multi core (and we would run same AI multi-core) Neon would be much faster, correct? In this case we would like to run on M1. 

Here is proposal let me know what you think:
a) we add an env variable to https://github.com/libxsmm/libxsmm/blob/main/src/libxsmm_cpuid_arm.c#L124 , e.g. libxsmm_cpuid_arm_m4_use_neon_non_gemm()
b) based on this value we set the eltwise kernels to either M1 or M4
c) for eltwise kernels (like gather/scatter) which don't have SSVE instruction support we bend them over to M1 as early as possible and we then can hopefully avoid some special ifs we are right now discussing.

stefan0re:
Sounds good to me. I am going to open a new PR with these features.

PR #950, opened 2025-03-13
Reference caller ppc64le

WillTrojak:
This adds the reference caller for ppc64le. 

The only issue at the moment are the sgemm and dgemm tests with `br gemm`  are failing. I'm not sure what this option is, is there a way to flag it as not supported? Or is it trivial to to support given we already have sgemm and dgemm? 

What do you think @alheinecke?


alheinecke:
Thanks @WillTrojak !

BRGEMM is our main kernel for AI. It's contraction over multiple A_i X B_i matmuls into a single C. It's essentially C[N][M] = \sum_i A_i[K][M] * B_i[N][K] in col-major. The "i" can be either provided as different A/B addresses array https://github.com/libxsmm/libxsmm/blob/main/samples/xgemm/gemm_kernel.c#L2569, as relative offset array https://github.com/libxsmm/libxsmm/blob/main/samples/xgemm/gemm_kernel.c#L2273 , or as fixed stride https://github.com/libxsmm/libxsmm/blob/main/samples/xgemm/gemm_kernel.c#L2360.

If you could add this, that would be great as it would allow to this version of libxsmm in Triton, Pytorch etc. for ppcle. If you feel comfortable, we first merge this one and then add BR later, this is also fine.

Just let me know and the I can get the review going or I wait for more changes.

alheinecke:
@WillTrojak : is this PR ready for review?

WillTrojak:
Yes, I think so.

alheinecke (approved):

PR #959, opened 2025-04-14
Fixed vector length initialization for nonunit lmul

siddharthrai:
Eltwise kernel was crashing in convolution tests due to uninitialized vector length for nonunit lmul.


alheinecke (approved):
LGTM

Issue #958, opened 2025-04-14
M4 SIMD execution


alheinecke:
@FreddieWitherden @stefan0re : @FreddieWitherden recently wondered if after your PR fsspmdm still works on M4. I assume when setting: LIBXSMM_AARCH64_USE_NEON=1 we get the behavior as before your PR?

Should we change the default and make the env variable LIBXSMM_AARCH64_M4_USE_SSVE?

FreddieWitherden:
> We can do that, but then the matrix multiplication would also run on NEON by default. Because at the moment, the execution of every primitive is decided here:
> 
> [libxsmm/src/libxsmm_cpuid_arm.c](https://github.com/libxsmm/libxsmm/blob/8071c0d80c3b439d48790f7f5ffe1c39db2b588a/src/libxsmm_cpuid_arm.c#L169-L173)
> 
> Lines 169 to 173 in [8071c0d](/libxsmm/libxsmm/commit/8071c0d80c3b439d48790f7f5ffe1c39db2b588a)
>  if (sme_supported == 1 && libxsmm_cpuid_arm_m4_use_neon_non_gemm() == 0) { 
>      result = LIBXSMM_AARCH64_APPL_M4; 
>  } else { 
>      result = LIBXSMM_AARCH64_APPL_M1; 
>  }

I think we might want to be able to have a different path taken for dense GEMM than for sparse GEMM.  What is the least invasive way of accomplishing this?

alheinecke:
What about forcing sparse GEMM to M1 / Aarch64 / Neon. The sparse code generator could just overwrite in the beginning to  M1 in case of M4. We do these demotions quite often in x86:
https://github.com/libxsmm/libxsmm/blob/main/src/libxsmm_main.c#L2193

FreddieWitherden:
This makes sense to me.  I do not think SME or the SVE variants it requires provide any advantage in the sparse cases (CSR or the a-sparse-reg stuff).

stefan0re:
At the moment there are no Sparse GEMM features that run on M1 / Aarch64, there are only some for NeoV1 and NeoV2. 
Or am I missing something?
[https://buildkite.com/intel/libxsmm-arm-jetson/builds/3308#0196b5bc-406f-4625-8eb1-98989233a5da](url)
https://buildkite.com/intel/libxsmm-arm-macmini/builds/2823#0196b5bc-614f-481a-8ec3-a1063682cef8

alheinecke:
[...]

Issue #957, opened 2025-04-12
How to generate SpMM kernel on M4


violet73:
Hi, I want to test the performance of SpMM kernel on M4. 

But I cannot figure out how to generate the kernel using libxsmm.

Many thanks for providing me some guidance!

alheinecke:
Can you please described what you are trying to do? Is it related to #958

violet73:
@alheinecke 
I just wonder how to generate SpMM kernel. 

I can see the way to generate the gemm kernel in the README.

Besides, I also want to see if libxsmm can use SME feature.

alheinecke:
Still not sure, what you are asking. LIBXSMM offers multiple sparse kernels, but not sparse * sparse, but dense * sparse = dense, or compact dense * sparse. However, we would need to know which examples you tried and into which issues you were running into.

violet73:
@alheinecke 

I want to use libxsmm to generate a kernel that computes $A \times B = C$, where $A$ is a sparse matrix and $B$ is a dense matrix. 

Specifically, $A$ is a sparse matrix that I read in from the matrix market format.

alheinecke:
Please have a look at #958, we have one specific kernel for ASIMD that has sparse support. However, more general spares ARM kernels are all targeting SVE and SVE2. For neither those (or even x86) we have matrixmarket readers. However, I think you should be able to write them quickly of the kernels we have suite you.

violet73:
Thank you for providing this information!

I seach for the related code according to #958 and find a sample that read a CSR format sparse matrix and call the corresponding [API](https://github.com/libxsmm/libxsmm/blob/c8af8f78e46386486313794b61306adf0aa25646/samples/pyfr/pyfr_driver_asp_reg.c#L241) to create SpMM kernel.

That's enough to implement what I want to do.

No further question before I start testing. Thank you again!

alheinecke:
great, then let me close this one and please submit PRs if needed/possible.

PR #964, opened 2025-05-20
Use NEON generator for spgemm on Mac M4

violet73:
The M4 CPU supports only the streaming mode SVE, which restricts the execution of SVE instructions to streaming mode exclusively. Current `libxsmm_generator_spgemm_csr_asparse_reg_aarch64_sve` generates SVE instructions without inserting the required `smstart` and `smstop` instructions, leading to illegal instruction errors. 

By manually adding `smstart` before the first SVE instruction and `smstop` after the last, most errors are avoided when running `test.sh` under `samples/pyfr`. However, several test cases still fail result verification.

The testing shows that the performance of SVE on the M4 CPU is lower than that of  NEON. The reason may be that the SVE is implemented very inefficiently on the SME coprocessor. As a result, using the NEON generator directly is a more efficient choice.


alheinecke:
I assume this is for addressing #958 ?

violet73:
Yes, based on #957 and #958, I test the `samples/pyfr` on M4 CPU and other regular SVE CPUs and found the above difference between them.

PR #968, opened 2025-06-02
printing BW in gemm_kernel tester

alheinecke:


samples/xgemm/gemm_kernel.c
```diff
@@ -2805,6 +2810,7 @@ int main(int argc, char* argv []) {
       }
       l_gemm_def.fuse_zpt_sub = 1;
     } else if (strcmp(argv[1], "MXFP4") == 0) {
+      l_dtype_a_size = 0.5;
```
egeor, line 2813: Strictly speaking this should be 0.53125 to account for the 8bit mxscale correct (which adds 1 byte per 32)?

egeor (approved):
Looks good to me!

Issue #970, opened 2025-06-13
New GEMM B^t Kernel for NEON


stefan0re:
I have implemented a new GEMM kernel for B^T in NEON, thereby I have also adapted the load and store register block function so that less vector registers are needed when m % 4 == 3.

![Image](https://github.com/user-attachments/assets/aebcc02a-ca99-4e03-891a-361baf34fcf2)

The benchmark results show that microkernel blocking with M = 12 outperforms M = 16 in most scenarios. However, the current RELU bitmask implementation is not compatible with M = 12, as it would require substantial changes to the bitmask logic.

How should this be handled and how important is the RELU bitmask currently?

alheinecke:
This is great and let's make sure we can get this merged!

For bitmask relu fusion we make sure LDC%8 == 0 (or even LDC%16 == 0), otherwise it should error out. At least this is what is happening for x86. Then we can access the ReLU mask in any case. I think it even says anywhere that the ReLU mask bits which are OOB are undefined. 

Unfortunately, I had no time to check the aarch64 code for this behavior, but I hope the explanation above provides enough leads to keep going...

BTW: I assume the better performance ASIMD (128bit code) for Grace in case of A and B non-transposed was incl. in your earlier PRs.

alheinecke:
Nope we have now trans and non-trans for Neoverse V2 at good perf-levels. Not sure about A^T and if we can do better here. But for now we can close #917 and #970

PR #971, opened 2025-06-18
New B^t GEMM for NEON

stefan0re:
This PR implements a better GEMM kernel for B^t matrices.
Additionally, the C load and store functions have been reworked so that fewer registers are needed for the C accumulator.
All changes apply only to NEON (arm).

#970


stefan0re:
The casting and the transposition parameter are now fixed.
I haven't found a nice way to remove the B transpose parameter either, so I did it with the enum.

alheinecke (approved):
LGTM

PR #985, opened 2025-08-18
Feature rv64 transpose

siddharthrai:
Added transpose kernel for RV64 supporting FP32, FP64, I32, and I64 data types.


siddharthrai:
> > Can you please merge main, such that we are conflict free and can run full CI?
> 
> Done.

alheinecke (changes requested):
The PR seems to introduce code which has not yet been fully developed (GEMM with A^T), or is it fully validated?

Some minor nit comment on the encoder.

src/generator_gemm.c
```diff
@@ -210,17 +210,18 @@ void libxsmm_generator_gemm_kernel( libxsmm_generated_code*        io_generated_
       return;
     }
 
+#if 0
```
alheinecke, line 213: This seems to be general section of the generator, I think we should revert

alheinecke (changes requested):
Looks good, just a minor NIT

src/generator_gemm.c
siddharthrai: This is a generic file, but the change is in the RV64 specific code. I think that reverting will make the TRANS_A case go to the reference code.

alheinecke (approved):

src/generator_gemm.c
alheinecke: OK, and I see you did now remove the dead code.

PR #997, opened 2025-12-06
AVX512 inline VNNI converts/shuffles for i8, i16 and bf16

alheinecke:
allowing flat/splat layouts is very important for simple tile based programming. This PR prepare for AVX10.2 adding flat/splat layout support for i8,i16,bf16 for AVX512. A future PR will add support for fp8/fp16 which will be available with AVX10.2.


egeor (approved):
Looks good to me!

PR #998, opened 2025-12-15
Feature sw pipeline panels

egeor:
This PR adds generalized SW pipelining for processing A/B matrices in AMX BRGEMMs, including panel-level processing to allow for transform reuse.


scripts/buildkite/libxsmm-all-tpp.yml
```diff
@@ -32,6 +32,13 @@ steps:
 #      CONFIG: gcc-122
     label: "SMMs SPR"
 
+  - command: "scripts/tool_test.sh samples/xgemm/kernel_test.sh"
```
alheinecke: Let's also add a GNR test

alheinecke (changes requested):
Let's make sure going back to random C matrices in the unit test doesn't break things.

alheinecke:
Can we also update the DMR testing with the commit?

alheinecke (approved):

PR #1005, opened 2026-01-08
Fixed obsolete macros from the encoder

siddharthrai:
Fixed riscv encoder tests to use latest encoder macros.


alheinecke (approved):
LGTM

PR #1009, opened 2026-02-13
Fixed BRGEMM execution for SME

stefan0re:
SME-BRGEMM is only possible when B is transposed and strides are used for reduction. 
The fallback to NEON isn’t pretty, but I didn’t find a better solution for now.

Minor change: added the A^T flag to the xgemm test output.


alheinecke (approved):
LGTM

PR #1014, opened 2026-06-01
Feature 8bit amx gemms via stack

egeor:
This PR enables 8bit AMX GEMMs via BF16 compute (including A and B xforms via SW pipeline)


alheinecke:
does this PR conflict with #1013, then we should merge #1013 first to get clean DMR support

alheinecke (approved):
LGTM

PR #1016, opened 2026-06-10
Fix BF8/HF8 NT fused-sigmoid GEMM: use 64-bit KMOVQ for reserved-mask save/restore in trans_KxN_8bit

egeor:
Fix BF8/HF8 NT fused-sigmoid GEMM: use 64-bit KMOVQ for reserved-mask save/restore in trans_KxN_8bit


alheinecke (approved):
lgtm

PR #1018, opened 2026-06-10
Improve VRCPPS with one Newton-Raphson iteration

egeor:
Improve VRCPPS with one Newton-Raphson iteration. This affected spradically avx2 gemms with fused sigmoid.


alheinecke (approved):
LGTM

PR #1019, opened 2026-06-10
Feature bf32 via amx gemms

egeor:
Enables BF32 on AMX via BF16 compute (simple F32->BF16 convert for inputs A and B).


alheinecke (approved):
LGTM

PR #1020, opened 2026-06-10
Enable unittests for CMake

Growl1234:
This PR enables the existing LIBXSMM tests through CMake/CTest.

The implementation is CMake-native: it builds the required test and sample executables directly and registers the default `tests/test.sh` entries as separate CTest tests, without invoking the legacy Makefile test target.

BLAS-dependent sample tests are kept optional and are enabled only when BLAS is available.

One additional modification: Use CMake-general `BUILD_SHARED_LIBS` option name instead of specific `XSMM_STATIC` to decide whether to build static or shared library.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -1,0 +1,75 @@
+name: LIBXSMM tests with CMake
+
+permissions:
+  contents: read
+  pull-requests: write
+
+on:
+  push:
+    branches: [ "main" ]
+  pull_request:
+    branches: [ "main" ]
+  workflow_dispatch: {}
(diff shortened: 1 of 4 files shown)


alheinecke:
I have to double check, but I thought I had fully removed the BLAS dependency, see #981

Growl1234:
> I have to double check, but I thought I had fully removed the BLAS dependency, see #981

I'll do some checks too, and I think I might need to revise the CMake configuration in #1015...

rengolin (approved):
This looks good to me, thanks! Let me see if I can get the tests going.

PR #1021, opened 2026-06-13
Extending GEMM's 6bit fields for Datatypes to eltwise and equation kernels

alheinecke:
The libxsmm_meltw_descriptor struct was refactored to replace three separate 1-byte datatype fields datatype datatype1, datatype2 with a single packed 32-bit datatypes field that stores up to five sub-fields (IN0, IN1, IN2, OUT, COMP) at 6 bits each, enabling a richer and more uniform representation of element-wise data types while keeping the overall descriptor size constant at 32 bytes.

To reclaim the byte needed for the increase of the datatype section by 1 byte, the standalone operation field, param field for elemetwise operations were merged into a single 16-bit param_operation field (operation in bits [2:0], param in bits [15:3]), and corresponding inline getter/setter helpers were introduced for both libxsmm_meltw_descriptor and libxsmm_gemm_descriptor.

All call sites across generator backends (x86, AArch64, RV64) were updated to use the new accessor functions and the LIBXSMM_MELTW_GETENUM_* / LIBXSMM_MEQN_GETENUM_* helpers for reading individual datatype sub-fields, removing the need for the intermediate internal_get_typesize_string helper.


egeor:
Looks good to me!

egeor (approved):

Issue #855, opened 2024-02-06
New Release Timeline


jeremylt:
#850 introduced some breaking changes. I was wondering if you folks are preparing for a release soon - a release would sure help libCEED communicate dependency requirements more easily and libXSMM is our recommended dependency for efficient CPU computation in our library.

Thanks!

alheinecke:
@jeremylt yes, we are cleaning up for a v2 release which breaks some of the API. We are getting close, I assume 2-3 months. You are right #850 got rid of the transitional _v2 names and made them default. 

We still want to incl some features smoothing rough edges, mainly kernels that return "NULL" today on certain arch (like SSE, BF16 for ASIMD, etc.). 

Perhaps some higher level API might still get removed, but the core functionality is now stable for v2.

jeremylt:
Awesome, thanks for the info!

mbanck:
Any update on this?

Also, it would be nice if your intermediate tags (like [1.old_kernelapi_rip](https://github.com/libxsmm/libxsmm/releases/tag/1.old_kernelapi_rip) wouldn't lexically sort above your latest stable release (1.17)

alheinecke:
We have still PPC64LE and system Z support pending. However, we are extensively CI/CD every commit on main and only PRs after review are merged. Is a version for you essential? Otherwise, please just pick head of main.

jeremylt:
It's definitely still the case that 
>  a release would sure help libCEED communicate dependency requirements more easily and libXSMM is our recommended dependency for efficient CPU computation in our library.

jeremylt:
Is libxsmm moving away from making any more releases? While that would not be my preference, it would be handy to know if that is libxsmm's plan so I can point our users at a specific commit hash as the dependency version instead of a version number

alheinecke:
you can use any main hash as these are heavily tested (hence main moves slowly). We expect v2 when we have ACE support: https://x86ecosystem.org/wp-content/uploads/2026/03/ACE-Whitepaper-v1.pdf

jeremylt:
Understood 

I get why years between releases is becoming normal for us scientific software developers with us all shretched so thin, but 'use LIBXSMM main and hope there isn't a bug in how we (libCEED) are using that dependency library' really isn't as comfortable of a thing for me to say to users as 'use version X of this dependency'.

boegel:
I agree that it's a shame that there hasn't been a new release of libxsmm since v1.17 which dates from Dec'21 (!)...

[...]

PR #1022, opened 2026-06-15
Improve GNU Make install layout for CMake and pkg-config

Growl1234:
Improves the GNU Make installation interface by generating and installing CMake package configuration files, so downstream projects can use `find_package(libxsmm CONFIG)` with GNU Make builds as well. The installed config exposes the imported target `libxsmm::libxsmm`, matching the CMake build path.

Also moves GNU Make generated pkg-config files from the library directory to the standard `lib/pkgconfig` location, and updates the module file accordingly.

Together, these changes can make the GNU Make and CMake installation layouts more consistent and easier for downstream build systems to consume.


Growl1234:
One related question: is there any tentative plan for a new release? I noticed that the latest release is still v1.17 from Dec 2021, while the main branch has accumulated many improvements since then. A newer release would be helpful for downstream packaging and for projects that prefer to depend on tagged versions rather than main-branch snapshots.

alheinecke:
We are discussing this in #855 and normally have new release with a major ISA update, in this case we are waiting for ACE.

alheinecke (approved):
LGTM

Makefile
```diff
@@ -1310,12 +1314,19 @@ $(OUTDIR)/libxsmmf-shared.pc: $(OUTDIR)/libxsmmf.$(DLIBEXT)
...
-	@ln -fs $(notdir $@) $(OUTDIR)/libxsmmf.pc
+	@ln -fs $(notdir $@) $(PPKGDIR)/libxsmmf.pc
   endif
 else
-.PHONY: $(OUTDIR)/libxsmmf-shared.pc
+.PHONY: $(PPKGDIR)/libxsmmf-shared.pc
 endif
 
+
+$(PCMKDIR)/libxsmmConfig.cmake: $(ROOTSCR)/libxsmmConfig.cmake $(ROOTSCR)/libxsmmConfigVersion.cmake.in $(PCMKDIR)/.make
```
alheinecke, line 1324: This is failing on MacOS, can you please have a look and submit a fix?

```
sed: 1: "./scripts/libxsmmConfig ...": invalid command code .
--
make: *** [lib/cmake/libxsmm/libxsmmConfig.cmake] Error 1
make: *** Waiting for unfinished jobs....
```

alheinecke (commented):
Please have a look on why this is failing on MacOS

Makefile
Growl1234: The input file was placed before the `-e` expression, which GNU sed accepts but BSD sed interprets as the sed program. I'll reorder the arguments to use the portable `sed -e EXPRESSION INPUT` form.

PR #1023, opened 2026-06-15
CMake: Build binaries and Fortran bindings

Growl1234:
With this PR, the corresponding generator executables (`libxsmm_binaryexport_generator` and `libxsmm_gemm_generator`) are built and installed explicitly in CMake, and the Fortran binding target (`libxsmm::libxsmmf`) is also generated, built, installed, and exported.


alheinecke (approved):
LGTM

PR #1026, opened 2026-06-18
Feature refactor datatypes

egeor:
This PR enables proper handling of sub-byte datatypes (instead of using the -- deprecated with this PR -- INTERPRETE_A_xxx


include/libxsmm_typedefs.h
```diff
@@ -481,46 +493,86 @@ typedef enum libxsmm_gemm_flags {
...
-  LIBXSMM_GEMM_FLAG_INTERPRETE_A_AS_INT4_VNNI2 = 2097152,
-  LIBXSMM_GEMM_FLAG_INTERPRETE_A_AS_INT4_VNNI8_INTLV = 4194304,
-  LIBXSMM_GEMM_FLAG_DECOMPRESS_A_VIA_BITMASK = 8388608,
-  LIBXSMM_GEMM_FLAG_INTERPRETE_A_AS_MXFP4_VNNI2 = 16777216,
-  LIBXSMM_GEMM_FLAG_USE_MxK_ZPT = 33554432,
-  LIBXSMM_GEMM_FLAG_USE_MxK_SCF = 67108864,
-  LIBXSMM_GEMM_FLAG_INTERPRETE_A_AS_MXFP4_VNNI8_INTLV = 134217728,
-  LIBXSMM_GEMM_FLAG_INTERPRETE_A_AS_INT2_VNNI4_INTLV = 268435456,
-  LIBXSMM_GEMM_FLAG_INTERPRETE_A_AS_INT1_VNNI4 = 536870912,
+  LIBXSMM_GEMM_FLAG_INTLV_A_FORMAT = 2097152,
```
alheinecke, line 496: since we have X2 datatypes (and will use them for A) I'm wondering if we can just derive INTLV when we use an X2 dataytype?

src/generator_common.c
```diff
@@ -902,7 +902,7 @@ void LIBXSMM_GEMM_SET_DESC_DATATYPE(libxsmm_datatype a_dt, libxsmm_datatype b_dt
   unsigned char uc_b = (unsigned char) b_dt;
   unsigned char uc_c = (unsigned char) c_dt;
   unsigned char uc_comp = (unsigned char) comp_dt;
-  unsigned char first = uc_a & 0xcf; /* 6 bits for a  */
+  unsigned char first = uc_a & 0x3f; /* 6 bits for a  */
```
alheinecke, line 905: cf is masking seems to be wrong this is 1100 1111 for the lower 6bits. Why?

src/generator_gemm_reference_impl.c
```diff
@@ -430,7 +430,7 @@ void libxsmm_setup_gemm_def(libxsmm_gemm_def* i_gemm_def, void *param, const lib
     l_gemm_def.unsigned_b = 1;
   }
 
-  if ((l_dtype_a == LIBXSMM_DATATYPE_I8) && ((i_xgemm_desc->flags & LIBXSMM_GEMM_FLAG_INTERPRETE_A_AS_MXFP4_VNNI2) > 0)) {
+  if ((l_dtype_a == LIBXSMM_DATATYPE_MXFP4X2) && ((i_xgemm_desc->flags & LIBXSMM_GEMM_FLAG_VNNI_A) > 0) && ((i_xgemm_desc->flags & LIBXSMM_GEMM_FLAG_INTLV_A_FORMAT) == 0)) {
```
alheinecke, line 433: OK.... this is connect with the abovem INTLV and X2 trigger different things, can you please quickly summarize here.

alheinecke (commented):
Just some minor clarifications.

include/libxsmm_typedefs.h
egeor: Nope, the X2 just signals that e.g. 4-bit datatypes are packed by 2 to form a byte. If there is vnni, interleave or both is orthogonal: see e.g. deprecated LIBXSMM_GEMM_FLAG_INTERPRETE_A_AS_MXFP4_VNNI2 and LIBXSMM_GEMM_FLAG_INTERPRETE_A_AS_MXFP4_VNNI8_INTLV. The interleave merely means that atop of the vnni formating we may have folding of the other dimension (as e.g. happens in the LIBXSMM_GEMM_FLAG_INTERPRETE_A_AS_MXFP4_VNNI8_INTLV format) -- see internal proposal text for refactoring.

src/generator_gemm_reference_impl.c
egeor: Yes, see comment above. The original composite flag is decomposed to the new ones. 1 for datatype, 1 for vnni and 1 for interleave.  So we are just carrying over the original information. In that sense, the X2 does not carry much information other than this dtype is guaranteed to have full bytes. -- see internal refactoring proposal text.

src/generator_common.c
egeor: I think this was a bug in main since it was not using the proper mask for the bits, right? It never manifested because we never went beyond the first 4 bits I believe.

alheinecke: oops you are right, I miss-read the change was the opposite. yepp this is a fix of main.

alheinecke (approved):
lgtm

PR #1032, opened 2026-06-21
Make BF32 a standard LIBXSMM datatype

alheinecke:
### Summary

This branch promotes `BF32` (TF32-style FP32 compute emulated on BF16 hardware) from an environment-variable-driven mode into a real `libxsmm_datatype`. Previously, BF32 GEMMs were expressed as plain FP32 descriptors that were re-routed to the bf16-emulated AMX path only when the `LIBXSMM_USE_FP32_VIA_BF16` env var was set. That implicit flag is replaced by an explicit `LIBXSMM_DATATYPE_BF32` value that flows through the code exactly like FP32 — same 4-byte storage — but carries the "compute via bf16" intent in the descriptor itself.

### Motivation

The env-var approach was global, process-wide, and invisible to the descriptor/JIT cache, making BF32 hard to request per-kernel and easy to misuse. Encoding it as a datatype makes the intent explicit at the API boundary and lets the normal dispatch/caching machinery distinguish BF32 from FP32.

### What changed

**New datatype**
- Added `LIBXSMM_DATATYPE_BF32` to the `libxsmm_datatype` enum (after `F32`), with typesize `4` and a passthrough in `LIBXSMM_GETENUM_SIGNED_DATATYPE` (libxsmm_typedefs.h).
- Mirrored the enum value into the Fortran bindings, shifting subsequent constants to stay aligned with C (libxsmm.f, libxsmm.f).
- Added `bf32` typename, `bf32f32` gemm-typename, and FP32-style `dvalue` handling (libxsmm_main.c).

**Removed the env-var flag**
- Deleted `libxsmm_cpuid_use_fp32_via_bf16()` and its declaration (libxsmm_cpuid_x86.c, libxsmm_cpuid.h). All `LIBXSMM_USE_FP32_VIA_BF16` reads are gone.

**Generators normalize on the datatype instead of the env var**
- generator_gemm.c: detect BF32 (`l_is_bf32_gemm`), normalize the working descriptor BF32→F32 so all generic processing is unchanged, gate the non-SPR fallback on the datatype, and re-stamp A/B→BF32 just before the AMX dispatch.
- generator_gemm_amx.c: drive the `fp32_via_bf16_sw_pipeline` from BF32 detection; the existing retag-to-BF16 block handles the rest.
- generator_gemm_reference_impl.c: the F32/F32/F32/F32 path now accepts BF32 A/B and applies the bf16 round-trip when A is BF32.

**Sample + tests**
- gemm_kernel.c: `BF32 BF32` selects the BF32 datatype directly (no `putenv`), with matching updates to random-matrix init, the gold reference compute, and the supported-precision check.
- generate_gemm_test_scripts.tpl: dropped the now-defunct env-var injection for `BF32_BF32_F32_F32`.

### Behavior

- BF32 descriptors route to the AMX bf16-emulated FP32 pipeline on SPR+; on non-AMX/non-SPR targets they fall back to the reference implementation (same as before, now keyed on the datatype rather than the env var).
- Plain FP32 is unaffected — it no longer gets silently diverted by a global env var.

### Validation

- Full library rebuild is clean under `-Werror`; the xgemm sample builds clean.
- BF32 runs across multiple shapes (16³, 32×48×64, 64³, 17×33×19) match the reference exactly (L2 rel.error 0, max error 0).
- FP32 regression unchanged.

### Notes / compatibility

- `libxsmm_cpuid_use_fp32_via_bf16()` is removed from the internal API (it was not part of the ABI surface). Any out-of-tree callers relying on `LIBXSMM_USE_FP32_VIA_BF16` should switch to requesting `LIBXSMM_DATATYPE_BF32`.
- The `libxsmm_datatype` enum values after `F32` shift by one; the Fortran constants were updated to match.


egeor (approved):
Looks good to me!

PR #1034, opened 2026-06-21
fix .gitignore for cmake, pkgconfig files in lib directory

alheinecke:


rengolin (commented):
can you not just ignore the whole `lib` dir?

egeor (approved):

rengolin (approved):

PR #1053, opened 2026-06-27
CMake: support Fortran-only package consumers

Growl1234:
This improves and makes safe the native CMake package integration for downstream Fortran projects.

Previously, the installed package configuration unconditionally called `find_dependency(Threads)`. Since CMake's `FindThreads` module requires either C or C++ to be enabled, this prevented a pure Fortran project from using:

```cmake
find_package(libxsmm CONFIG REQUIRED)
```

The thread usage requirements are now propagated to the exported LIBXSMM targets directly, so the installed package no longer depends on `Threads::Threads` or invokes `FindThreads` during package discovery.

This also adds support for the optional Fortran component:

```cmake
find_package(libxsmm CONFIG REQUIRED COMPONENTS Fortran)
target_link_libraries(my_target PRIVATE libxsmm::libxsmmf)
```

The component is reported as unavailable when the installed LIBXSMM build does not contain the Fortran interface.

P.S. The README is updated to document the native CMake build, installation, and downstream package usage. GNU Make remains documented as the reference build path, particularly for its broader build-time customization and legacy sample coverage.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,6 @@
 cmake_minimum_required(VERSION 3.13 FATAL_ERROR)
 
-set(LIBXSMM_VERSION "1.17")
+set(LIBXSMM_VERSION "1.17.0")
 
 project(libxsmm
   VERSION ${LIBXSMM_VERSION}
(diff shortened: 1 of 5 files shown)


alheinecke (approved):
LGTM

PR #1063, opened 2026-06-29
CMake: Derive version from version.txt

Growl1234:


Growl1234:
This might be too much for the PR (small patch), but in fact I am curious whether `version.txt` can be replaced since it can be changed into something like `main-XXXX` automatically.

alheinecke (approved):
LGTM

Growl1234:
> I am curious whether `version.txt` can be replaced since it can be changed into something like `main-XXXX` automatically.

Or, revise `tool_version.sh` so it keeps the existing version information in `version.txt` when no tag is fetched in the fork.

PR #1064, opened 2026-06-29
Version 2.0.0

alheinecke:


README.md
egeor, line 1: @alheinecke What about the i2 and i2 datatypes?

alheinecke: just add was is missing and commit

egeor (approved):

PR #1065, opened 2026-06-29
Revise version parsing logic

Growl1234:
`version.txt` is automatically refreshed from the Git state and therefore represents development metadata rather than a stable release version. In particular, a checkout in fork without fetching tags can produce values such as `main-3857`, which might be interpreted as release version `3857.0.0`.

This PR separates the two kinds of information:

* Add a canonical `VERSION` file for the release/API version.
* Use `VERSION` consistently for CMake, Make, generated headers, interfaces, pkg-config metadata, and package compatibility information.
* Retain `version.txt` for the branch name and development revision only, and ignore when exporting release package; for GNU Makefile, only parse it when exists (for develop/`git clone`).
* Make `tool_version.sh` always include the release version when refreshing `version.txt`, including in tag-less or shallow checkouts.


alheinecke:
@Growl1234 let me know when done, and I will review

Growl1234:
Thank you! I will re-check again to see if I'm missing anything.

Since both `version.txt` and `VERSION` are used in current PR, which may cause somewhat duplicate (though `version.txt` now relies on `VERSION`), I think one thing that better make clear is: what does original `version.txt` really mean and whether such a string `<branch>-<version>-<patch>` is important for maintainers. This could help us decide whether to go on or to consider current status as ready for review.

alheinecke:
The ```<branch>-<version>-<patch>``` is important when one doesn't have the .git folders anymore for internal branch check-out. It allows to unique identify a commit. So we should keep this.

Growl1234:
Thanks for making it clear. So I would consider this PR as ready.

alheinecke (approved):
LGTM

PR #1066, opened 2026-07-10
Add Transpose-A and strided BRGEMM support for ARM SME

stefan0re:
Previously, only Transpose B was supported, and not all BRGEMM variants worked with the strided (`strdbr`) address mode. This PR adds Transpose A support and enables the strided BRGEMM path for the remaining cases, so both are now available for SME. Additionally, beta = 0 is now supported.


alheinecke:
Thanks! Let me create a branch such that we can test it. Under scripts/buildkite/ is an ARM script which runs ArmIE, can we also add SME tests here?

alheinecke (commented):
LGTM, let's merge into the branch and look at CI

PR #1071, opened 2026-07-20
Addig ppc64le support

WillTrojak:
This is a reopening of the powerpc support PR. It adds:

* NN support for FP32 and FP64, with the various batchings. 
* Reference trampoline for the unsupported data formats and other options.
* Sparse A matrix support for FP32/FP64 PyFR matrices

The should support little endian: power8, power9, power10 and power11 features.


alheinecke:
@WillTrojak Thanks!

what's the difference compared to the feature_powerpc branch? It has some minor conflicts. In order to test before merging into main we have 2 options:
a) merge this one into feature_powerpc_new; are we going to loose something with is the current feature_powerpc branch?
b) rebase this one on top of feature_powerpc
What would you recommend?

alheinecke:
@WillTrojak please let me know what you think on the questions above. We might release 2.1.0 sooner than expected due to aarch64 packaging issue in Fedora... would be cool to get this into 2.1.0 as well.

WillTrojak:
There are some fixes in this branch. I don't think I'm able to push to the libxsmm/feature_powerpc branch. You could try merging into the feature_powerpc branch, but it might be easier to just use this branch as it's recently rebased off main and works.

PR #1075, opened 2026-07-22
Add aarch64 CMake test

Growl1234:
To keep an eye on CMake workflow on aarch64 architecture.

The flag `-march=armv8.1-a` is passed to CMake (consistent with GNU Makefile).


alheinecke (approved):
LGTM

PR #1077, opened 2026-07-22
Install public headers under `include/libxsmm`

Growl1234:
This PR installs LIBXSMM headers under `${prefix}/include/libxsmm` instead of directly under `${prefix}/include`. The exported include directories are updated accordingly for both GNU Makefile and CMake installations.

Existing downstream source code can continue to use:

```c
#include <libxsmm.h>
```

because `${prefix}/include/libxsmm` is exported as the include search path.


alheinecke (approved):
LGTM

PR #1079, opened 2026-07-24
CMake: Adjust Fortran module and header installation

Growl1234:
- Seperate the installation dir of Fortran modules via using `CMAKE_INSTALL_Fortran_MODULES`. Fedora doesn't install `.mod` files under `/usr/include/.../`, but under `/usr/usr/lib64/gfortran/modules/.../`. This change allows to pass `-DCMAKE_INSTALL_Fortran_MODULES:PATH=%{_fmoddir}/%{name}` to CMake in RPM spec. The default would still be `LIBXSMM_INSTALL_INCLUDEDIR`, convenient for building and installation oneself.
- Drop installation of `libxs_source.h`. It's belong to header-only mode, which is not really implemented in CMake and moreover discouraged by downstream packaging.


alheinecke:
Please indicate when ready to merge

alheinecke (approved):
LGTM

PR #1082, opened 2026-07-26
CMake: Various little tweaks

Growl1234:
- Include documentation installation.
- Rename export targets and config templates to kebab-case (`<PackageName>Config.cmake` -> `<lower-case-name>-config.cmake`) for wider compatibility: `find_package(LIBXSMM)` and `find_package(libxsmm)` would be both valid (previously only `find_package(libxsmm)`).
- Clean up redundant `LIBXSMM_VERSION` assignment in config template as CMake will handle it automatically.
- Add CMake build type settings; set default as `RelWithDebInfo`.


alheinecke (approved):
LGTM

PR #1084, opened 2026-07-30
Introduce Packit CI to track RPM building

Growl1234:
This PR brings the maintenance of the Fedora RPM build into the upstream repository, with the aim of strengthening the connection and collaboration between upstream development and downstream packaging. It also makes it possible to track the potential impact of upstream changes on the Fedora RPM during development. The goal is to minimize the maintenance burden associated with the RPM build and keep the upstream project more closely synchronized with the version available in the Fedora repositories.

The relevant CI tests are handled by Packit. Every commit pushed to the `main` branch triggers Packit CI, and users with triage permissions can also trigger a build from a pull request by commenting `/packit build`. This requires maintainers to install [Packit-as-a-Service app](https://github.com/marketplace/packit-as-a-service) for this repository and obtain access to Packit (see https://packit.dev/docs/guide).


.packit.yaml
```diff
@@ -0,0 +1,46 @@
...
+jobs:
+  - &copr_build
+    job: copr_build
+    trigger: release
+  - <<: *copr_build
+    trigger: commit
+    branch: main
+  - <<: *copr_build
+    trigger: pull_request
+    manual_trigger: true
```
LecrisUT: I do not recommend `manual_trigger`. It was added to cp2k because they didn't want it otherwise, and it has generally caused more trouble than anything.

.packit.yaml
```diff
@@ -0,0 +1,46 @@
...
+
+  - job: koji_build
+    trigger: commit
+    dist_git_branches:
+      - fedora-all
+
+  - job: bodhi_update
+    trigger: commit
+    dist_git_branches:
+      - fedora-branched
```
LecrisUT: I would drop these until downstream gives the blessing. Will check with Dominick (second admin on Fedora there) for help figuring out the longer term ownership.

.packit.yaml
```diff
@@ -0,0 +1,46 @@
...
+  - fedora-latest-aarch64
+  - fedora-latest-ppc64le
+
+jobs:
+  - &copr_build
+    job: copr_build
+    trigger: release
+  - <<: *copr_build
+    trigger: commit
+    branch: main
```
LecrisUT, line 30: Unless you have a persistent copr repo these would not help. You can use `preserve_project` to make it persistent, or add it to a bigger project.

alheinecke (approved):
as there were no more comments from fedora maintainers, approved.

Growl1234:
@alheinecke To make the workflow valid, we need the follow-up workflow:

> This requires maintainers to install [Packit-as-a-Service app](https://github.com/marketplace/packit-as-a-service) for this repository and obtain access to Packit (see https://packit.dev/docs/guide).

I wonder how you think about this 🤔

alheinecke:
Is this all you need: https://github.com/packit/notifications/issues/747

Growl1234:
Yes, it's exactly what we need to make Packit work. Thank you!

PR #1086, opened 2026-08-06
Fix fsspmdm performance on Apple M4.

FreddieWitherden:
This ensures that we pick the correct vector length for Apple chips that support SME.


alheinecke (approved):
LGTM

PR #1087, opened 2026-08-15
Add MXINT8 GEMM datatype (for ACE) and harden the JIT-failure path

alheinecke:
### Summary

This branch adds a new MX datatype, **`LIBXSMM_DATATYPE_MXINT8`** (MX-scaled signed `int8` with an E8M0 block scale, block size 32), including a native AMX-class code path and an architecture-portable reference implementation. It also fixes a NULL-pointer dereference that made applications crash instead of reporting an unsupported GEMM configuration.

### 1. `MXINT8` datatype

**Format.** A and B are signed `int8` with a shared E8M0 scale per block of 32 elements; C and compute are `F32`. Structurally it mirrors pure `MXBF8` — only the tile opcode and the numeric decode differ.

**Numerics.** Each `MXINT8` element carries an implicit exponent bias of `-6`, so a product of two operands accumulates $`-12`$ independent of tensor rank. With both block scales as $2^{s-127}$ the accumulator adjustment is

$`\text{exp\_adjust} = -12 + s_A + s_B - 254`$

The reference implementation evaluates this in double precision to avoid spurious underflow.

**Backend.** Native codegen is emitted only where the hardware supports it (`TOP4MXBSSPS`); every other target falls back to the reference implementation. The existing `MXBF8`/`MXHF8` detector was generalized and renamed `libxsmm_x86_is_Amx8_Bmx8_gemm` (from `is_Amxfp8_Bmxfp8_gemm`) since it now also covers `MXINT8` — the old name was actively misleading.

**Design note.** `AB_COMMON_PREC` deliberately returns `MXINT8` for `A = B = MXINT8` rather than canonicalizing to `MXBF8`. Canonicalizing would have been a smaller diff, but it hides the real type from every downstream consumer. As a consequence `MXINT8` is listed explicitly at each `AB_COMMON == MXBF8` gate, and `libxsmm_cpuid_dot_pack_factor()` had to be extended so the k-loop blocking factor stays at 4.

**Files touched:** `libxsmm_typedefs.h` (enum, `TYPESIZE`, `GETENUM_SIGNED`), `generator_gemm.c`, `src/generator_gemm_common.{c,h}`, `generator_gemm_avx512_microkernel.c`, `generator_gemm_sse_avx_avx2_avx512.c`, `generator_gemm_reference_impl.c`, `libxsmm_cpuid_x86.c`, `libxsmm_main.c`, plus `README.md` (datatype list).

**Constraints.** `MXINT8` GEMM requires A in VNNI and B in VNNI-and-transposed layout; other layouts are rejected.

### 2. Graceful handling when no kernel can be produced

Some configurations are rejectable by both the JIT and the reference path — e.g. `MXINT8 x MXINT8` without `AVNNI=1 BVNNI=1 BTRANS=1`. In that case dispatch correctly returns `NULL`, but the caller crashed before it could report anything.

**Root cause (library):** `libxsmm_get_kernel_info()` in `libxsmm_main.c` read `xinfo->is_reference_kernel` *before* its own `NULL != xinfo` guard, so `libxsmm_get_kernel_info(NULL, ...)` segfaulted. The assignment now lives inside the guard.

**Sample (`gemm_kernel.c`):**
- `jit_matmul()` tests the dispatched pointer for `NULL` *before* querying kernel info, prints an explicit message that neither a JIT-ed kernel nor a reference fallback was available, and exits with `EXIT_FAILURE` instead of faulting.
- A new `print_offending_cmdline()` echoes the failing invocation so it can be replayed directly. In explicit-size mode it prints `argv` verbatim. In **file-input mode** — where sizes come from a size file and are therefore absent from the command line — it reconstructs the *equivalent explicit-size invocation*: the current `M N K LDA LDB LDC` are substituted, the size-file argument and `run_check` are dropped, and `nopf` is inserted for the prefetch argument that only the explicit form carries. The two forms are byte-identical for the same case.

Example output:

```
ERROR: LIBXSMM could not create a kernel for this GEMM configuration:
       neither a JIT-ed kernel nor a reference (fallback) implementation is available.
       Please run with LIBXSMM_VERBOSE=-1 and/or with a debug mode LIBXSMM library for details.
Offending command line:
  ./gemm_kernel MXINT8 MXINT8 F32 F32 64 64 64 64 64 64 1 0 0 0 0 0 0 0 0 nopf nobr 1 0 1 0
```


egeor (approved):
Looks good to me!

PR #1089, opened 2026-08-23
Gate signal handler installation on verbosity

### LINKED ISSUE #1088: LIBXSMM constructor overrides SIGSEGV/SIGABRT handlers before the host runtime initializes
LIBXSMM currently installs handlers for `SIGABRT` and `SIGSEGV` from its constructor (`libxsmm_init`). This can interfere with signal handling established by the host runtime.

I encountered this while debugging CP2K built with Intel LLVM 2026.1.1. Although CP2K was compiled with `-g -traceback` and contains the `.trace` section, a segmentation fault only resulted in:

```text
Segmentation fault (core dumped)
```

rather than the expected Intel Fortran runtime traceback:

```text
forrtl: severe (174): SIGSEGV, segmentation fault occurred
Image              PC                Routine            Line        Source
...
```

Tracing signal setup with

```bash
strace -f -k -e trace=rt_sigaction ...
```

shows that LIBXSMM installs its `SIGSEGV` handler from its ELF constructor, before the Intel Fortran runtime is initialized:

```text
rt_sigaction(SIGSEGV, {sa_handler=0x..., ...}, ...)
 > libc.so.6(__libc_sigaction+...)
 > libc.so.6(ssignal+...)
 > libcp2k.so(libxsmm_init+0x24d)
 > ld-linux-x86-64.so.2(call_init+...)
 > ld-linux-x86-64.so.2(_dl_init+...)
```

Later, when the Intel Fortran runtime initializes, it only queries the existing handler:

```text
rt_sigaction(SIGSEGV, NULL,
             {sa_handler=<LIBXSMM handler>, ...}, ...)
 > cp2k.psmp(for_rtl_init_+0x2fd)
 > cp2k.psmp(main+0x17)
```

and consequently does not install its own `SIGSEGV` handler.

The resulting failure path is approximately:

```text
real SIGSEGV
  -> LIBXSMM signal handler
  -> restore previous handler
  -> raise(SIGSEGV)
  -> previous handler / SIG_DFL
  -> core dump
```

This is not specific to the original fault itself: in this case the underlying crash is an unrelated Intel MPI / `mpi_f08` issue. LIBXSMM only changes how that failure is diagnosed.

More generally, I think a library constructor unconditionally taking ownership of process-wide fatal signals such as `SIGSEGV` and `SIGABRT` can be problematic for applications embedding LIBXSMM. It may interfere with Fortran runtimes, sanitizers, crash reporters, debuggers, or application-provided signal handling.

There is also a related question about performing library finalization from a fatal-signal handler. Such finalization may involve operations for which async-signal-safety cannot generally be assumed.

Would it be possible to reconsider this behavior? Possible approaches might include:

* not installing fatal-signal handlers by default;
* providing a compile-time or runtime opt-out for LIBXSMM signal handling;
* making signal handling explicitly opt-in for applications that want this behavior;
* or otherwise avoiding interference with host/runtime signal diagnostics.

For comparison, removing the LIBXSMM/LIBXS `SIGSEGV` registrations allows the Intel Fortran runtime to install its own handler and restores the expected traceback.

Environment where this was observed:

```text
CP2K master branch
Intel oneAPI toolkit 2026.1.1
LIBXSMM 2.1.0
Linux x86_64
```

LIBXS has a closely related implementation and independently installs the same fatal-signal handlers, so I will report the corresponding issue there as well.

Growl1234:
Only install signal handlers when `LIBXSMM_VERBOSE` is set (or `LIBXSMM_SIGNAL` overrides).

Copy of https://github.com/hfp/libxs/commit/a450161ef4c7f892f730a0aad09c861c47d6a8ab. Fix #1088.


alheinecke (changes requested):
Code change looks good to me (incl. the change in the library)!

Can you please document the change under https://github.com/libxsmm/libxsmm#runtime-control in the README.md? (mainly a summary of #1088). Then we can merge.

alheinecke (approved):
lgtm

PR #1090, opened 2026-08-28
GEMM: bound JIT stack scratch and operand size instead of crashing (#992)

### LINKED ISSUE #992: Segfault on Larger GEMM Sizes
### Summary
I attempted to use `libxsmm_dispatch_brgemm` for BF16 workloads but encountered a **segmentation fault (core dumped)** for several large problem sizes during a single-instance run.
Smaller problem sizes execute correctly without any issues (e.g., m=128, k=512, n=1024).
A minimal reproducer and system details are provided below.

---

### Environment
* **LIBXSMM version:** feature_print_bw-1.17-3780 (commit 25693892)
* **LIBXSMM lib:** libxsmm.a (not inclusing any other lib)
* **Architecture:** x86_64  

---

### Problem sizes that segfault

| `m` | `k` | `n` |
|----:|----:|----:|
| 43 | 768 | 50265 |
| 86 | 768 | 30522 |
| 173 | 768 | 30522 |
| 86 | 768 | 50257 |
| 173 | 512 | 30522 |
| 2048 | 512 | 250112 |
| 4096 | 768 | 50272 |
| 8192 | 768 | 50005 |
| 8192 | 768 | 50265 |
| 4096 | 512 | 32128 |
| 256 | 768 | 20005 |
| 256 | 512 | 197951 |

(Other shapes run fine.)

---

### Minimal reproducer

```cpp
#include <libxsmm.h>
#include <iostream>

void run(size_t M, size_t N, size_t K,
         const libxsmm_bfloat16* A,
         const libxsmm_bfloat16* B,
         libxsmm_bfloat16*       C,
         unsigned int            batch,
         size_t strideA,
         size_t strideB)
{
  libxsmm_gemm_shape shape{};
  libxsmm_gemm_batch_reduce_config br{};
  libxsmm_bitfield flags = LIBXSMM_GEMM_FLAGS('N', 'N');

  shape.m   = N;   // note: swapped on purpose
  shape.n   = M;
  shape.k   = K;
  shape.lda = N;
  shape.ldb = K;
  shape.ldc = N;

  shape.a_in_type = LIBXSMM_DATATYPE_BF16;
  shape.b_in_type = LIBXSMM_DATATYPE_BF16;
  shape.out_type  = LIBXSMM_DATATYPE_BF16;
  shape.comp_type = LIBXSMM_DATATYPE_F32;

  br.br_type          = LIBXSMM_GEMM_BATCH_REDUCE_STRIDE;
  br.br_stride_a_hint = strideB * sizeof(libxsmm_bfloat16);
  br.br_stride_b_hint = strideA * sizeof(libxsmm_bfloat16);
  br.br_unroll_hint   = 0;

  libxsmm_gemmfunction ker =
      libxsmm_dispatch_brgemm(shape, flags, 0 /*prefetch*/, br);
  if (!ker) { std::cerr << "JIT compilation failed\n"; return; }

  libxsmm_gemm_param p{};
  p.a.primary   = const_cast<libxsmm_bfloat16*>(B);
  p.b.primary   = const_cast<libxsmm_bfloat16*>(A);
  p.c.primary   = C;
  p.op.tertiary = &batch;

  ker(&p);    // <-- Segmentation fault for shapes listed above
}

```
---
### Observations and Questions:

* **LIBXSMM_MAX_MNK Variable**
Can we use the LIBXSMM_MAX_MNK variable to set an upper bound for the problem size?
I noticed that (M * N * K)^(1/3) <= 64 is considered the optimal size for LIBXSMM, but larger sizes still execute successfully.

* **Upper Bound for Problem Sizes**
Is there a recommended upper bound for M, N, and K to avoid segmentation faults?

alheinecke:
### Problem

`libxsmm_dispatch_brgemm` returns a valid kernel for shapes far outside LIBXSMM's
intended small-GEMM range, but calling that kernel segfaults — e.g. `m=50265, n=43,
k=768` (BF16, flat A, stride-BRGEMM). Smaller shapes work, which made the failure
look size-dependent but arbitrary.

### Root cause

Not an arithmetic overflow (the instruction encoder is already 64-bit safe: 
`libxsmm_x86_instruction_alu_imm` falls back to a `MOVABS` + temp-register sequence 
for immediates beyond int32). The real problem is that the JIT'ed prologue performs an 
**unbounded `sub rsp, imm`** whose size scales with the problem dimensions.

All size-dependent stack allocations live in
`libxsmm_generator_gemm_setup_stack_frame_allocate_scratch` (and its AArch64/RV64
clones):

| Term | Trigger |
|---|---|
| `m*n*4` | `vnni_format_C` |
| `m*k*dt`, `k*n*dt` | `*_stack_alloc_tensors` (flat A converted to VNNI on the stack) |
| `max(m*k, k*n)*4 + m*n*4` | FP8 stack tensors |
| `4*n*max(ldb,k)` | B SW-pipeline |
| `m*k*2` | AMX panel SW-pipeline |

The reporter's case (`'N','N'`, BF16, A not in VNNI) sets
`avnni_gemm_stack_alloc_tensors`, giving `a_size = m*k*2` = **77 MB** on an 8 MB
stack. Every shape in the reported failure table exceeds 8 MB; every shape reported
as working stays below it.

### Changes

**Stack-scratch budget (the actual fix)**

* New `libxsmm_generator_gemm_stack_scratch_{limit,size,exceeded}()` in
  `generator_gemm_common.c`. `..._size()` models every term above in `long long`,
  tracking the A-side and B-side reformat buffers separately (on non-ACE targets only
  the A buffer is actually allocated, so summing both would reject working shapes).
* The gate runs at the top of all three allocators (x86, AArch64, RV64) and raises
  `LIBXSMM_ERR_UNSUP_SIZE`. The existing generic fallback in `libxsmm_main.c` then
  dispatches `libxsmm_generator_gemm_reference_kernel`, so the shape still computes
  correctly — it is simply not JIT-ed.
* Default budget is 2 MB, overridable via `LIBXSMM_GEMM_STACK_SCRATCH_LIMIT`
  (Bytes; `0` restores the previous unchecked behavior). A warning naming the shape,
  the required size and the env var is printed at `LIBXSMM_VERBOSE>=2`.

**Operand-size bound**

* `libxsmm_generator_gemm_shape_fits_int32()` runs first in
  `libxsmm_generator_gemm_kernel` and rejects any GEMM whose A, B or C operand reaches
  2 GiB. x86 memory operands carry a signed 32-bit displacement, so this limit cannot
  be lifted by up-casting the generator's arithmetic — it needs a gate.

**Diagnostics, test, docs**

* Added the missing error string for `LIBXSMM_ERR_UNSUP_SIZE` (it previously printed
  "an unknown error occurred (error #90053)").
* `tests/gemmbig.c`: drives four shapes derived from the issue through
  `libxsmm_dispatch_brgemm` and validates the result against a naive reference.
  Registered in `tests/CMakeLists.txt`; the GNU Makefile auto-discovers it.
* `documentation/libxsmm_mm.md`: new "Problem Size Limits" section covering both
  limits, the env var, the transparent reference fallback, and a clarification that
  `LIBXSMM_MAX_MNK` only guides the BLAS-style auto-dispatch wrappers — it has no
  effect on `libxsmm_dispatch_gemm` / `libxsmm_dispatch_brgemm` (a question raised in
  the issue).

### Validation

* All shapes from #992 now run to completion. Setting
  `LIBXSMM_GEMM_STACK_SCRATCH_LIMIT=0` reproduces the original segfault, confirming
  the gate is what fixes it.
* `samples/xgemm/kernel_test.sh`: 3600 tasks, exit 0, no regressions.
* Threshold calibration (BF16, flat A): `1024x1024x1024` and `2048x16x512` — both
  exactly at the 2 MB budget — still JIT natively; `2048x1024x1024` (4 MB) and larger
  fall back. Those were genuinely unsafe stack allocations, not a lost optimization.

### Notes / follow-up

* BRGEMM variants multiply the scratch by the **runtime** batch count
  (`imul temp, a_size` / `sub rsp, temp`), which cannot be bounded at JIT time. The
  per-instance size is now capped, but a very large batch count can still exhaust the
  stack.
* Several genuine 32-bit products remain in `generator_gemm_amx.c` (2641/2743/2797
  `(lda*m)*(4/a_vnni_adjustment)`, 4373/4444/4477 `(long long)(m_adj*lda)`, 5866) plus
  `extra.nflops = 2*m*n*k` in `libxsmm_main.c`. They are now unreachable for the
  shapes that used to crash, but should be up-cast in a follow-up.

Fixes #992


src/generator_gemm_common.h
```diff
@@ -246,6 +246,30 @@ void libxsmm_generator_gemm_setup_stack_frame_fill_ext_gemm_stack_vars( libxsmm_
     libxsmm_micro_kernel_config*        i_micro_kernel_config,
     const libxsmm_gp_reg_mapping*       i_gp_reg_mapping );
 
+/**
```
egeor: Nit: move the define at the top of the file?

egeor (approved):
Just a cosmetic nit, otherwise looks good to me!

src/generator_gemm_common.h
alheinecke: got catch, I missed this in my diff on the cmd.

PR #1094, opened 2026-08-29
Fix fused (br)GEMM bias add without binary col-bcast flag (#766)

### LINKED ISSUE #766: Erroring out when requesting fused (br)gemm with bias addition and we do not have binary col bcast flag set

alheinecke:
### Problem

The fused GEMM/BRGEMM ("EXT ABI") path only ever implements a **column-broadcast**
bias add. In `libxsmm_generator_gemm_setup_fusion_microkernel_properties()`
(`src/generator_gemm_common.c`) the `fused_*colbias` micro-kernel properties are set
*only* if `LIBXSMM_MELTW_FLAG_BINARY_BCAST_COL_IN_0` or `..._IN_1` is present.

When a caller (e.g. tpp-mlir's `tpp.fused_brgemm`) requests
`LIBXSMM_MELTW_TYPE_BINARY_ADD` with a full `MxN` bias and no broadcast flag, the
post-op is **silently dropped**: dispatch succeeds and the kernel returns wrong
results. The reference implementation in `src/generator_gemm_reference_impl.c` had
the same restriction, so the reference JIT fallback did not rescue it either.

Reproduced with a minimal `libxsmm_dispatch_brgemm_ext` program
(`M=32, N=16, K=32, br=2`, all-ones `A`/`B`, bias `= 100`):

| config | expected `C[0]` | before |
| --- | --- | --- |
| `BCAST_COL_IN_0` | `164.0` | `164.0` |
| no bcast flag | `164.0` | `64.0` (bias ignored, no error) |

### Changes

1. **`src/generator_gemm.c`** — reject the unsupported configuration up front.
   If the descriptor carries a C-side binary meltw op that is not `ADD` with a
   `BCAST_COL_IN_0`/`IN_1` flag, raise `LIBXSMM_ERR_INVALID_GEMM_CONFIG` instead of
   silently ignoring the requested post-op. This is the "error out" behaviour the
   issue asks for, and it gives frontends running with
   `LIBXSMM_DISABLE_REFERENCE_GEMM_FALLBACK` a clean `NULL` dispatch so they can
   split the ops themselves.

2. **`src/generator_gemm_reference_impl.c`** — teach the reference path the
   non-broadcast case, so the existing reference fallback computes the correct
   result instead of dropping the bias:
   - new `fuse_bias_add_nobcast` and `ldd` fields in `libxsmm_gemm_def`,
     derived from `meltw_flags` / `meltw_ldx`;
   - `libxsmm_ref_apply_preop()` now selects `UNARY_NONE` / `BINARY_NONE` with the
     descriptor-provided `ldd` instead of the hard-coded `BCAST_COL` flag and
     `ldi = m`. The broadcast path is unchanged.

Only the C-side binary post-op is affected; the unary post-ops (`RELU`, `SIGMOID`,
`VNNI` conversion) and the A/B-side pre-ops are untouched. The matrix-equation path
is unaffected — it already decomposes correctly when the flag is absent.

### Validation

- New behaviour verified for the reported case: `bcast=1` (JIT) and `bcast=0`
  (reference) both yield `164.0`, across `beta=1` / `LIBXSMM_GEMM_FLAG_BETA_0`
  and with `ldd != M` (bias with padded leading dimension).
- Matrix-equation regression check: `relu(add(brgemm(A,B), bias))` with and
  without `BCAST_COL_IN_1` dispatches and runs identically to before.
- `samples/xgemm/kernel_test_srf.sh`: **174/174 tasks pass**, no new failures.
- Clean build (`make`, `-Werror`), no new warnings.

Fixes #766.


egeor (approved):
Looks good to me!

PR #1096, opened 2026-08-30
Cleanup code for colbias fusion (#818)

### LINKED ISSUE #818: Cleanup code for colbias fusion
[CLEANUP] Consolidate bias fusion variable check (e.g. here https://github.com/libxsmm/libxsmm/blob/main_stable/src/generator_gemm_amx.c#L406) and specialize base on the dtype of the input bias

alheinecke:
## Cleanup code for colbias fusion

Closes #818.

@egeor : let me know if this was the issue you wanted to address!!

### Background

@egeor filed #818 in Sep 2023 asking to *"consolidate bias fusion variable check and specialize based on the dtype of the input bias"*, pointing at the bias-pointer advancement in `generator_gemm_amx.c`.

The colbias fusion state was encoded as five mutually-exclusive booleans in `libxsmm_micro_kernel_config`:

```c
unsigned int fused_bcolbias;   /* BF16 */
unsigned int fused_hcolbias;   /* F16  */
unsigned int fused_b8colbias;  /* BF8  */
unsigned int fused_h8colbias;  /* HF8  */
unsigned int fused_scolbias;   /* F32  */
```

all derived from the single descriptor field `meltw_datatype_aux`. Every consumer then had to re-derive the bias datatype from those flags. Since the issue was filed, BF8/HF8 were added, so the duplication grew: **29** five-way OR chains and **3** hand-written ternary cascades reconstructing the dtype from the flags.

### Change

Replace the five flags with one flag plus the datatype:

```c
unsigned int     fused_colbias;
libxsmm_datatype colbias_dtype;  /* only meaningful when fused_colbias != 0 */
```

plus `LIBXSMM_GEMM_FUSED_COLBIAS(config, dtype)` for the genuinely dtype-specific sites. Both fields are set once, in `libxsmm_generator_gemm_setup_fusion_microkernel_properties()`.

Consequences:

- 29 five-way OR chains collapse to `fused_colbias == 1`.
- The per-dtype pointer advancements (the exact sites the issue pointed at) become a single block using `LIBXSMM_TYPESIZE(colbias_dtype)`.
- The three ternary cascades in `generator_gemm_amx.c` are replaced by one helper, `libxsmm_generator_gemm_amx_colbias_load_props()`.
- The mixed-FP8 stack-emulation path no longer has to reconstruct the bias FP8 flavor by hand; it just sets `colbias_dtype = F32` after the upconvert. (That reconstruction was the source of an earlier numerical bug on mixed A/B FP8 GEMMs.)

Net **−87** lines across `generator_common.h`, `generator_gemm_common.c`, `generator_gemm_amx.c`, `generator_common_aarch64.c`, `generator_gemm_aarch64.c`, `generator_gemm_rv64.c`.

Paths that do not support the full dtype set keep an explicit narrowed predicate so behavior is unchanged: AArch64/RV64 handle BF16 and F32 only, and `libxsmm_generator_gemm_store_C_ace()` has no F16 bias support.

### Second commit: fix an uninitialized read on AArch64

The refactor initially regressed `sgemm_nt_eqld_bop0_uop3_cvnni0` on `LIBXSMM_TARGET=arm_v82` (check-norm `0.218` instead of `0.0`). Note `bop0` means *no colbias at all*, so the colbias logic was not at fault.

Root cause is a pre-existing use of uninitialized memory in `libxsmm_generator_gemm_aarch64_get_initial_m_blocking()`:

```c
if( ((i_xgemm_desc->flags & LIBXSMM_GEMM_FLAG_TRANS_AB) == 2) && io_micro_kernel_config->fused_relu_nobitmask == 1){
```

That function only calls `libxsmm_generator_gemm_init_micro_kernel_config_aarch64()` (which `memset`s and populates the struct) **at its end**. Its first caller is `libxsmm_generator_gemm_aarch64_setup_n_blocking()` from `libxsmm_generator_gemm_aarch64_kernel()`, where `l_micro_kernel_config` is still a raw stack variable — so `fused_relu_nobitmask` is stack garbage.

The fallout:

1. Garbage → wrong `m_blocking` → wrong `n_blocking`.
2. `m_blocking` is recomputed later, when the struct *is* valid, and gets the correct value.
3. `n_blocking` and `m_blocking` then disagree, so `l_vec_reg_acc_start = vector_reg_count - (n_blocking * m_total_blocks)` allocates overlapping accumulator registers → wrong results.

The guard `(flags & LIBXSMM_GEMM_FLAG_TRANS_AB) == 2` is "B transposed", which is why only `_nt_` tests are sensitive. Shrinking the struct (5 `unsigned int` → 1 `unsigned int` + 1 enum) moved `fused_relu_nobitmask` to a different offset where it aliased different garbage — the refactor exposed the bug rather than causing it.

Fix: initialize the config before the first blocking query. I deliberately did not move the init to the top of `get_initial_m_blocking`, because it `memset`s and the caller sets `avnni_*_stack_alloc_tensors` before the later call. RV64 does not have this pattern.

### Validation

**x86 — generated code is unchanged.** 1463 JIT kernels dumped across `hsw/skx/clx/cpx/spr/gnr/dmr/srf` × 8 precision combinations × beta 0/1 × four unary post-ops × `nobr`/`strdbr` × aligned and masked M. Byte-identical to the previous implementation once embedded reference-kernel function pointers are normalized. A native `srf` numerical sweep over the same matrix (4352 output lines) shows zero differences.

**AArch64 — generated code is unchanged.** Because `libxsmm_generator_gemm_kernel()` is arch-agnostic, kernels can be cross-generated from an x86 host. 25600 configurations over `v81/v82/sve128/neov2/sve256/a64fx/sve512/appl_m4` × 5 precision combinations × M=64,37 × beta × transA/transB × 5 colbias dtypes (including none) × 4 unary post-ops × `nobr`/`strdbr`; 9760 produced real code, **0** byte differences.

**AArch64 — executed on Grace (Neoverse-V2).** Deterministic A/B against the pre-refactor build, 8 BRGEMM variants of the failing configuration:

| target | before the AArch64 fix | after |
|---|---|---|
| `arm_v82` baseline | 0/8 fail | 0/8 fail |
| `arm_v82` this PR | 8/8 fail | **0/8 fail** |
| `sve128` baseline / this PR | 0/8 fail | 0/8 fail |

`sve256` fails on baseline and this branch alike (Grace is SVE128) and is unrelated.

### Notes for reviewers

- The AArch64 fix changes generated code for `*_nt_*` AArch64 GEMMs, since blocking previously depended on undefined behavior. This is intended — it makes the choice deterministic and consistent with the later recomputation.
- Worth a follow-up: `libxsmm_generator_gemm_store_C_ace()` derives the bias load width from the **C** datatype rather than the bias datatype. That is latent breakage if the two ever differ; I left it alone here to keep this PR behavior-preserving.


src/generator_common.h
```diff
@@ -1700,6 +1697,10 @@ LIBXSMM_EXTERN_C typedef struct libxsmm_micro_kernel_config {
   unsigned int ace_b_prefetch_ctrl;
 } libxsmm_micro_kernel_config;
 
+/* colbias fusion is requested and the bias operand has datatype DTYPE */
```
egeor: Nit: Move to top of file

egeor (approved):
This is the cleanup I had in mind -- looks good to me. Just a comsetic nit in a separate comment.

src/generator_common.h
alheinecke: fixed

PR #1097, opened 2026-08-30
apply Newton iterations for reciprocal/rsqrt on ASIMD (#774)

### LINKED ISSUE #774: Eltwise unary reciprocal and rsqrt require Newton iteration on ASIMD
Eltwise unary reciprocal and rsqrt require Newton iteration to meet reasonable accuracy with ASIMD. Accuracy is reasonable for the SVE based code-path (Newton iteration).

alheinecke:
The ASIMD code-path emitted only FRSQRTE for eltwise unary rsqrt, i.e. a raw estimate with ~5-8% relative error, while the SVE path already refined the estimate with Newton iterations. This exposed a large accuracy gap between the two AArch64 uArch code-paths.

Refine the FRSQRTE estimate with FMUL/FRSQRTS/FMUL Newton steps, using the same iteration count as SVE (3 for fp32, 7 for fp64). The required temporary registers were already reserved for this operation on both ISAs.

Also let the ASIMD reciprocal honor LIBXSMM_ULP_PRECISION: emit a plain FRECPE when an estimate is requested, otherwise keep FRECPE+FRECPS+FMUL. This makes LIBXSMM_ULP_PRECISION the single knob for trading accuracy for speed on ASIMD, matching the SVE behavior.

Applied to both the mateltwise unary and the matrix-equation register-block generators.

Fixes #774


egeor (approved):
QQ: Should we use the analogous of libxsmm_cpuid_x86_use_high_prec_eltwise_approx function (e.g. as we did for some reciprocal in the past) ?

alheinecke:
harmonizing between aarch64 and x86 is needed, filed a new issue on this #1101

PR #1099, opened 2026-08-31
Fix Error out in equations when shape inference is not possible (#905)

### LINKED ISSUE #905: Error out in equations when shape inference is not possible
In equation cases where shape inference is not possible, error out with proper message

alheinecke:
### Problem
Equations accepted TPP nodes with no valid shape-inference rule. Instead of failing at
dispatch, LIBXSMM built the kernel with wrong or undersized temporaries, causing
out-of-bounds accesses at execution time. Some generator paths already detected
unsupported constructs but only printed to `stderr` and returned mid-generation,
leaving a broken yet apparently valid kernel.

### Solution
Validate the tree when the equation is finalized and fail via the standard mechanism:
`libxsmm_dispatch_meqn()` returns `NULL` with a specific diagnostic.

Now rejected:
- reduce opcodes without `REDUCE_ROWS`/`REDUCE_COLS`
- padding transforms (`PAD{M,N,NM}_MOD{2,4}`, `NORM_TO_VNNI{2,4,8}_PAD`)
- root-only opcodes as inner nodes (`UNZIP`, `SCATTER`, `REPLICATE_COL_VAR`, `RELU` with bitmask)
- `GATHER`/`REDUCE_COLS_IDX` whose input is not a user-provided argument
- binary/ternary operands not covering the output extent (broadcast flags honored)
- arguments with non-positive `M`/`N`

### Also fixed: `samples/equation/equation_matmul`
Built by CTest but never executed, and had rotted into a guaranteed segfault:
- `tmp` was allocated in bytes with the wrong dimension (reference kernels use it as an
  `m_C x n_D` float buffer)
- `arg_singular_attr` was only partially initialized

### Validation
- all 9 `samples/equation/equation_test/*.sh` pass, no existing equation is rejected
- `equation_matmul` F32: passes 3 shapes x 3 fusion modes, valgrind-clean

### Known issue (pre-existing, not addressed)
`equation_matmul` BF16 mode: equations 3/4 (`*_BRGEMM_A_VNNI`) produce `inf` and eqn4
writes past the BF16 output buffer. Belongs in the BF16 BRGEMM codegen; track separately.

Fixes #905


src/libxsmm_matrixeqn.c
```diff
@@ -865,6 +865,195 @@ int libxsmm_meqn_is_binary_opcode_reduce_to_scalar (unsigned int opcode) {
...
+       (opcode == LIBXSMM_MELTW_TYPE_UNARY_SCATTER)            ||
+       (opcode == LIBXSMM_MELTW_TYPE_UNARY_REPLICATE_COL_VAR)  ||
+       ((opcode == LIBXSMM_MELTW_TYPE_UNARY_RELU) && ((flags & LIBXSMM_MELTW_FLAG_UNARY_BITMASK_2BYTEMULT) > 0)) ) {
+    result = 1;
+  }
+  return result;
+}
+
+/* An operand is only usable if it covers the extent of the op's output in all non-broadcast dimensions. */
+LIBXSMM_API_INTERN int libxsmm_meqn_operand_shape_is_valid( libxsmm_meqn_bcast_type bcast_type,
```
egeor: move to header?

egeor (approved):
Minor nit wrt function declaration

src/libxsmm_matrixeqn.c
alheinecke: fixed

PR #1100, opened 2026-09-01
Makefile: Separate pkg-config build and install paths

### LINKED ISSUE #1083: Install directory inconsistent
https://github.com/libxsmm/libxsmm/pull/1077 this PR introduced a bug

When including LIBXSMM built locally but not installed, you need to use the include directory `libxsmm/include`, but `libxsmm.pc` reports that you should use `libxsmm/include/libxsmm`, which does not exist.

Growl1234:
Fix #1083.


Makefile
```diff
@@ -1129,6 +1129,14 @@ endif
 	@echo "LIBXSMM installing pkg-config, CMake config, and module files..."
 	@$(MKDIR) -p $(PREFIX)/$(PPKGDIR)
 	@$(CP) -va $(PPKGDIR)/*.pc $(PREFIX)/$(PPKGDIR) 2>/dev/null || true
+	@for pc in $(PPKGDIR)/*-static.pc $(PPKGDIR)/*-shared.pc; do \
+		if [ -f "$$pc" ]; then \
+			$(SED) -e "s|^prefix=.*|prefix=$(ALIAS_PREFIX)|" \
```
alheinecke, line 1134: Change overall looks good to me, first set of tests is running. The SED command looks like it's compatible with GNU sed under MAC. Have you tried this under MAC (so we catch it before merge and CI in main turning red?)

alheinecke (commented):
minor question, otherwise it looks good.

Makefile
Growl1234: Sorry I didn't test on MacOS as I don't have a Mac PC.

alheinecke: OK, the command looks fine for both BSD and GNU sed, so let's hope it goes through

alheinecke (approved):
lgtm

